PR #24, opened 2020-06-24
[ODS-4299] Cleanup remaining bulk references

semalaiappan:


semalaiappan:
removed logistics\scripts\modules\build-management.psm1: lines 190 and 191, sets up rules for bulk on new channels

PR #27, opened 2020-06-26
[ODS-4046] Recreating student, staff or parent after deletion returns 400 Bad Request

gmcelhanon:
This PR implements configuration file-based cache expiration settings for the Descriptors cache and Person UniqueId/USI cache.

See also https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/44

### DIFF
--- a/Application/EdFi.Ods.WebApi/Web.Base.config
+++ b/Application/EdFi.Ods.WebApi/Web.Base.config
@@ -30,6 +30,7 @@
     <add key="uniqueIdValidation:featureIsEnabled" value="false" />
     <add key="caching:personUniqueIdToUsi:slidingExpirationSeconds" value="14400" />
     <add key="caching:personUniqueIdToUsi:absoluteExpirationSeconds" value="0" />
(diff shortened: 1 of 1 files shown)


Application/EdFi.Ods.WebApi/Web.Base.config
```diff
@@ -30,7 +30,7 @@
     <add key="uniqueIdValidation:featureIsEnabled" value="false" />
     <add key="caching:personUniqueIdToUsi:slidingExpirationSeconds" value="14400" />
     <add key="caching:personUniqueIdToUsi:absoluteExpirationSeconds" value="0" />
```
blmeyers, line 32: @vimayya The default absolute expiration being used here for the person cache is infinite.  This means a default installation will still face the issue described in the linked ticket.  Is the expectation that we should have a reasonable default here that will still allow expiration of the cache or that we will communicate through docs to work around the issue using this flag?

blmeyers: I think we mostly covered this in standup, so we'll just need to make sure docs are up to date.

PR #43, opened 2020-07-13
[ODS-4334] Download smo from nuget

tbmeador-dlp:


logistics/scripts/modules/database/database-management.psm1
```diff
@@ -3,25 +3,31 @@
...
-        Add-Type -Path $dllPath | Out-Null
-        Write-Verbose "Loaded $assembly from $dllPath"
+& "$PSScriptRoot\..\load-path-resolver.ps1"
+Import-Module -Force -Scope Global (Get-RepositoryResolvedPath 'logistics\scripts\modules\database\database-utility.psm1')
+
+function Test-SqlServerModuleInstalled { $null -ne (Get-InstalledModule | where -Property Name -eq SqlServer) }
+
+function Test-SqlServerModuleImported { $null -ne (Get-Module SqlServer) }
+
+function Install-SqlServerModule {
```
blmeyers, line 13: Does it make more sense to use the verb "Use" here since you're also importing it?  https://docs.microsoft.com/en-us/dotnet/api/System.Management.Automation.VerbsOther.Use?view=powershellsdk-1.1.0  Could probably go either way, but curious on your opinion.

tbmeador-dlp: `Use` might make more sense here, still waiting on `Ensure` https://github.com/PowerShell/PowerShell/issues/11710. As I work more in powershell I'm finding that I dislike their verb restrictions. We end up in situations like this where we're having to look up which verb is considered correct. Something like `Ensure-SqlServerModuleIsLoaded` would be easier to understand but doesn't fit within the existing restrictions.

logistics/scripts/modules/database/database-management.psm1
```diff
@@ -3,25 +3,31 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
```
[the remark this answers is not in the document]
tbmeador-dlp: done

PR #47, opened 2020-07-17
[ODS-4321] Update Authorization Strategies for TPDM Aggregates

gmcelhanon:
Paired with https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/70.


Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6178 @@
...
+        @claimId AS INT,
+        @claimName AS nvarchar(max),
+        @parentResourceClaimId AS INT,
+        @existingParentResourceClaimId AS INT,
+        @claimSetId AS INT,
+        @claimSetName AS nvarchar(max),
+        @authorizationStrategyId AS INT,
+        @msg AS nvarchar(max),
+        -- Well-known values
+        @createActionId AS INT = 1,
```
jamessmckay, line 14: cant make this assumption due to deployments can be different. need to set these values using a lookup.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6178 @@
...
+
+    DECLARE @claimIdStack AS TABLE (Id INT IDENTITY, ResourceClaimId INT)
+
+    SELECT @applicationId = ApplicationId
+    FROM [dbo].[Applications] WHERE ApplicationName = 'Ed-Fi ODS API';
+
+    BEGIN TRANSACTION
+
+    -- Push claimId to the stack
+    INSERT INTO @claimIdStack (ResourceClaimId) VALUES (@claimId)
```
jamessmckay, line 27: i am confused here, how is this temporary table being set?

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6178 @@
...
+                WHERE ResourceClaimId = @claimId
+            END
+        END
+  
+    -- Processing claim sets for http://ed-fi.org/ods/identity/claims/domains/tpdm
+    ----------------------------------------------------------------------------------------------------------------------------
+    -- Claim set: 'Ed-Fi Sandbox'
+    ----------------------------------------------------------------------------------------------------------------------------
+    SET @claimSetName = 'Ed-Fi Sandbox'
+    SET @claimSetId = NULL
```
jamessmckay, line 70: this is redundant. this is a valid assumption that the sandbox claim set will be available.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6178 @@
...
+
+    SELECT @claimSetId = ClaimSetId
+    FROM dbo.ClaimSets
+    WHERE ClaimSetName = @claimSetName
+
+    IF @claimSetId IS NULL
+    BEGIN
+        PRINT 'Creating new claim set: ' + @claimSetName
+
+        INSERT INTO dbo.ClaimSets(ClaimSetName, Application_ApplicationId)
```
jamessmckay, line 80: not sure i following here, from what i can tell, we are duplicating the claim set sandbox into the claim set sandbox?

jamessmckay: The more i think about this, i think we can make the assumption that the sandbox claim set exists and we can remove this code.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6178 @@
...
+        VALUES (@claimSetName, @applicationId)
+
+        SET @claimSetId = SCOPE_IDENTITY()
+    END
+  
+    PRINT 'Deleting existing actions for claim set ''' + @claimSetName + ''' (claimSetId=' + CONVERT(nvarchar, @claimSetId) + ') on resource claim ''' + @claimName + '''.'
```
jamessmckay, line 86: concerned here, we are deleting the sandbox claim sets, and than recreating them. should the realignment be done with tpdm?

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6178 @@
...
+    DELETE FROM dbo.ClaimSetResourceClaims
+    WHERE ClaimSet_ClaimSetId = @claimSetId AND ResourceClaim_ResourceClaimId = @claimId
+    
+
+    -- Claim set-specific Create authorization
+    SET @authorizationStrategyId = NULL
+    
+
+    IF @authorizationStrategyId IS NULL
```
jamessmckay, line 95: this appears to be duplicated

Application/EdFi.Ods.Extensions.TPDM/Artifacts/PgSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,487 +1,5254 @@
...
+    _existingParentResourceClaimId INTEGER;
+    _claimSetId INTEGER;
+    _claimSetName VARCHAR(255);
+    _authorizationStrategyId INTEGER;
+    -- Well-known values
+    _createActionId INTEGER := 1;
+    _readActionId INTEGER := 2;
+    _updateActionId INTEGER := 3;
+    _deleteActionId INTEGER := 4;
+    _claimIdStack INTEGER ARRAY;
```
jamessmckay, line 17: i know we talked about the naming convention, it should be snake case... eg application_id

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: It's a table variable, declared at the top:
```sql
DECLARE @claimIdStack AS TABLE (Id INT IDENTITY, ResourceClaimId INT)
```

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: It just adds unnecessary complexity to the XSLT style sheet to eliminate it. It's ultimately a question of what we want to optimize for. I'm putting forth a direction of moving away from manual maintenance of SQL scripts, so this redundancy wouldn't be much of a concern (other than possibly due to the "bloat" of the script) and keeping the style sheet simpler is better, so long as it is not causing any problems.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: I'm going to agree with [your earlier comment](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pull/47#discussion_r457371518) and make the change to Actions to perform look ups, and suggest we leave these look ups alone.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: For actions (whether they are on the resource claim as the default authorization metadata, or for a resource / claim set combination as overrides of the defaults), the actions for the resource / claim set are recreated entirely.

Nothing outside the scope of the resource (for default authorization metadata), or the resource / claim set combination is affected.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: As I replied above, this is to be expected from a generated artifact. I'm proposing a move towards eliminating manual authoring of SQL scripts as the primary point of maintenance of this security metadata.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: I think that with these values in particular that people would have no reason to introduce additional actions or to reassign their values. However, this is a database table out in the wild with an identity column, so anything is possible. I will make the change to the scripts to look up these action Ids.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/PgSql/Data/Security/1010-TPDM-ResourceClaims.sql
gmcelhanon: Done.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+    -- Claim set: 'Ed-Fi Sandbox'
+    ----------------------------------------------------------------------------------------------------------------------------
+    SET @claimSetName = 'Ed-Fi Sandbox'
+    SET @claimSetId = NULL
+
+    SELECT @claimSetId = ClaimSetId
+    FROM dbo.ClaimSets
+    WHERE ClaimSetName = @claimSetName
+
+    IF @claimSetId IS NULL
```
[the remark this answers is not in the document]
gmcelhanon: Fair points, all, for the longer term. These were all decisions I made for the use case(s) at hand -- making it easier to manage the metadata and generate a script that would reproduce it in the database.

The XML extract from the database that I've produced is the version that I would want at runtime in the API. However we decide to break that down for management purposes in the future, at this point, I think that's the desired final runtime product ... and the TPDM document matches that structure right now for a more direct iterative authoring experience.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
--- SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
gmcelhanon: Yep, after about the 2nd time of preserving the first 6 lines of the script, I should have just added that to the XSLT template. I'll add it to both files.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+
+                UPDATE dbo.ResourceClaims
+                SET ParentResourceClaimId = @parentResourceClaimId
+                WHERE ResourceClaimId = @claimId
+            END
+        END
+  
+    -- Setting default authorization metadata
+    PRINT 'Deleting default action authorizations for resource claim ''' + @claimName + ''' (claimId=' + CONVERT(nvarchar, @claimId) + ').'
+    DELETE FROM dbo.ResourceClaimAuthorizationMetadatas
```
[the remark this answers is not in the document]
gmcelhanon: Yes, the "leave existing data alone" behavior is intentional. If you are wanting to add resource claims below an existing claim, I wanted to be able to tap into that Ed-Fi Standard hierarchy without redefining all it's security metadata. There's always more than 1 way to skin a cat, but I was going for the simpler direct route here.

If you want to clear default authorization metadata, just include the element as empty. If you look at the XSLT, the DELETE portion of logic is generated by the presence of that element, and then the INSERT statements are generated with each `<Action>` element. So, an empty `<DefaultAuthorization>` element will result in the deletion of those actions.
```xslt
<xsl:template match="DefaultAuthorization" mode="SqlServer">
    -- Setting default authorization metadata
    PRINT 'Deleting default action authorizations for resource claim ''' + @claimName + ''' (claimId=' + CONVERT(nvarchar, @claimId) + ').'
    DELETE FROM dbo.ResourceClaimAuthorizationMetadatas
    WHERE ResourceClaim_ResourceClaimId = @claimId
    <xsl:apply-templates select="Action" mode="SqlServer-DefaultAuthorization" />
</xsl:template>
```
This behavior supports the authoring of _change_ scripts on top of already existing metadata, allowing new resource claims to be added to the existing hierarchy without disturbing or redefining existing authorization metadata, while still giving you enough control to completely redefine it if you need to by clearing out unnecessary authorization metadata on different resource claims. The Actions under a particular ClaimSet for a resource works the same way:
```xslt
<xsl:template match="Actions" mode="SqlServer-ClaimSet">
    PRINT 'Deleting existing actions for claim set ''' + @claimSetName + ''' (claimSetId=' + CONVERT(nvarchar, @claimSetId) + ') on resource claim ''' + @claimName + '''.'
    DELETE FROM dbo.ClaimSetResourceClaims
    WHERE ClaimSet_ClaimSetId = @claimSetId AND ResourceClaim_ResourceClaimId = @claimId
    <xsl:apply-templates select="Action" mode="SqlServer-ClaimSet" />
</xsl:template>
```
Side Note: In looking at that just now, I see an inconsistency in that I've got an `<Actions>` containing element under the `<ClaimSet>` but not the `<DefaultAuthorization>` element which has direct `<Action>` elements as children. That's something we'll want to align at some point.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+            PRINT 'Creating new claim: ' + @claimName
+
+            INSERT INTO dbo.ResourceClaims(DisplayName, ResourceName, ClaimName, ParentResourceClaimId, Application_ApplicationId)
+            VALUES ('tpdm', 'tpdm', 'http://ed-fi.org/ods/identity/claims/domains/tpdm', @parentResourceClaimId, @applicationId)
+
+            SET @claimId = SCOPE_IDENTITY()
+        END
+    ELSE
+        BEGIN
+            IF @parentResourceClaimId != @existingParentResourceClaimId
```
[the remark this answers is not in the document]
gmcelhanon: This is a good catch. I will correct this logic and regenerate the scripts.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+    SELECT @readActionId = ActionId
+    FROM [dbo].[Actions] WHERE ActionName = 'Read';
+
+    SELECT @updateActionId = ActionId
+    FROM [dbo].[Actions] WHERE ActionName = 'Update';
+
+    SELECT @deleteActionId = ActionId
+    FROM [dbo].[Actions] WHERE ActionName = 'Delete';
+
+    BEGIN TRANSACTION
```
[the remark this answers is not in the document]
gmcelhanon: This is here because during development I wanted a transaction around the changes so that I could re-extract the final XML metadata to do inspection before rolling back the transaction for the next iteration. I just never thought to remove it, but I will remove explicit transaction management from the final SQL Server scripts here (Postgres apparently does an implicit transaction around the whole pgsql block, from what I could tell).

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+    IF @authorizationStrategyId IS NULL
+    BEGIN
+        SET @msg = 'AuthorizationStrategy does not exist: ''Relationships with Education Organizations only''';
+        THROW 50000, @msg, 1
+    END
+
+    INSERT INTO dbo.ResourceClaimAuthorizationMetadatas(ResourceClaim_ResourceClaimId, Action_ActionId, AuthorizationStrategy_AuthorizationStrategyId)
+    VALUES (@claimId, @DeleteActionId, @authorizationStrategyId)
+
+    -- Processing claim sets for http://ed-fi.org/ods/identity/claims/tpdm/teacherPreparationProviderProgram
```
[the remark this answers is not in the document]
gmcelhanon: I thought this was strange too, but here's the back story.

I was seeing an error in the TPDM template log where it indicated that the API client was missing a claim, but the message didn't provide an indication of what claim set was actually in use. So the changes in the Ed-Fi-ODS repository for this ticket enhance the error message so that it now includes the API client's assigned Claim Set.

As it turns out, this resource is being loaded under that "Bootstrap" claim set, as the following improved error message indicates:
```
EdFi.Ods.Api.Services.Controllers.TPDM.TeacherPreparationProviderPrograms.TeacherPreparationProviderProgramsController: 2020-07-18 19:42:50,936 [13] ERROR EdFi.Ods.Api.Services.Controllers.TPDM.TeacherPreparationProviderPrograms.Teacher
EdFi.Ods.Common.Security.EdFiSecurityException: Access to the resource could not be authorized. Are you missing a claim? This resource can be authorized by the following claims:
    http://ed-fi.org/ods/identity/claims/tpdm/teacherPreparationProviderProgram
    http://ed-fi.org/ods/identity/claims/domains/tpdm/noFurtherAuthorizationRequiredData
    http://ed-fi.org/ods/identity/claims/domains/tpdm
The API client has been assigned the 'Bootstrap Descriptors and EdOrgs' claim set with the following resource claims:
    http://ed-fi.org/ods/identity/claims/domains/systemDescriptors
    http://ed-fi.org/ods/identity/claims/domains/managedDescriptors
    http://ed-fi.org/ods/identity/claims/domains/educationOrganizations
    http://ed-fi.org/ods/identity/claims/accountabilityRating
    http://ed-fi.org/ods/identity/claims/classPeriod
    http://ed-fi.org/ods/identity/claims/communityOrganization
    http://ed-fi.org/ods/identity/claims/communityProvider
    http://ed-fi.org/ods/identity/claims/communityProviderLicense
    http://ed-fi.org/ods/identity/claims/course
    http://ed-fi.org/ods/identity/claims/educationOrganizationNetwork
    http://ed-fi.org/ods/identity/claims/educationOrganizationNetworkAssociation
    http://ed-fi.org/ods/identity/claims/educationOrganizationPeerAssociation
    http://ed-fi.org/ods/identity/claims/educationServiceCenter
    http://ed-fi.org/ods/identity/claims/feederSchoolAssociation
    http://ed-fi.org/ods/identity/claims/learningObjective
    http://ed-fi.org/ods/identity/claims/learningStandard
    http://ed-fi.org/ods/identity/claims/learningStandardEquivalenceAssociation
    http://ed-fi.org/ods/identity/claims/localEducationAgency
    http://ed-fi.org/ods/identity/claims/location
    http://ed-fi.org/ods/identity/claims/postSecondaryInstitution
    http://ed-fi.org/ods/identity/claims/program
    http://ed-fi.org/ods/identity/claims/school
    http://ed-fi.org/ods/identity/claims/stateEducationAgency
   at EdFi.Ods.Security.Authorization.EdFiAuthorizationProvider.GetAuthorizationDetails(EdFiAuthorizationContext authorizationContext)
   at EdFi.Ods.Security.Authorization.EdFiAuthorizationProvider.<AuthorizeSingleItemAsync>d__7.MoveNext() in D:\buildagent\work\7540bc60aa3a490f\Ed-Fi-ODS\Application\EdFi.Ods.Security\Authorization\EdFiAuthorizationProvider.cs:line 100
```
So, that's why the Create permission has been added for the Bootstrap claim set -- because the TPDM load process needed it to be.

If you look at the security metadata that was in place before my work, you'll find that _teacherPreparationProviderProgram_ resource claim was actually located under _educationOrganizations_ (along with a bunch of other non-EdOrg resources, see [ODS-4380](https://tracker.ed-fi.org/browse/ODS-4380)) and with explicit default authorization metadata using "No Further Auth" for all operations (which I don't think is actually used because there is an override for the "Bootstrap" claim set on the Create operation on educationOrganizations that takes precedence). In any case, the existing metadata has been cleaned up, but the claim set in use by the TPDM load process necessitated this particular setting.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+            BEGIN
+                PRINT 'Repointing claim ''' + @claimName + ''' (ResourceClaimId=' + CONVERT(nvarchar, @claimId) + ') to new parent (ResourceClaimId=' + CONVERT(nvarchar, @parentResourceClaimId) + ')'
+
+                UPDATE dbo.ResourceClaims
+                SET ParentResourceClaimId = @parentResourceClaimId
+                WHERE ResourceClaimId = @claimId
+            END
+        END
+  
+    -- Processing claim sets for http://ed-fi.org/ods/identity/claims/domains/tpdm
```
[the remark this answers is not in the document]
gmcelhanon: That statement is sort of true, but I covered it in my earlier response. It depends on how the script being produced is inteded to be used. The use case I started with was producing an additional script to fix up (rather than completely replace) the existing metadata. Even with a full replacement of the 1010 TPDM security metadata script, there's still the need to reference the existing Ed-Fi security metadata (particularly for expanding on the existing claims hierarchy) without fear of overwriting what's there (or inadvertently changing it if it had been customized). That was my approach here.

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
```diff
@@ -1,465 +1,6004 @@
...
+
+            SET @claimId = SCOPE_IDENTITY()
+        END
+    ELSE
+        BEGIN
+            IF @parentResourceClaimId != @existingParentResourceClaimId
+            BEGIN
+                PRINT 'Repointing claim ''' + @claimName + ''' (ResourceClaimId=' + CONVERT(nvarchar, @claimId) + ') to new parent (ResourceClaimId=' + CONVERT(nvarchar, @parentResourceClaimId) + ')'
+
+
```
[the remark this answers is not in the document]
gmcelhanon: I decided to include this as a seam for someone to drop TPDM-specific authorization metadata on. My thought was, if survey was in use with and without TPDM-aware API clients, would the same permissions apply to both or would it be helpful to someone if they had an already existing claim to be drop some permissions on rather than having to carve it out themselves later. So I leaned on the err of introducing claims at those seams for that purpose.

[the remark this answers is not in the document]
gmcelhanon: It was specifically called out in the analysis by the TPDM team under the [Credentials](https://techdocs.ed-fi.org/pages/viewpage.action?spaceKey=EDFIODS&title=TPDMDEV-243+-+Review+API+authorizations+strategies#TPDMDEV243ReviewAPIauthorizationsstrategies-Credentials).

Looking at it, I think that I started by adding new claim set to it (and subsequently realized that we just needed to continue using just the Bootstrap and Ed-Fi Sandbox claim sets). I think because of the null check problem you called out above, that I didn't notice in the end state that the metadata here would be changing its parent, and that once the claim set metadata was removed there was no longer a reason to keep it in here (it will revert to being a child of _relationshipBasedData_ and will using Relationship-based authorization).

I will remove this from the TPDM Security Metadata XML file and after the fix to the XSLT stylesheet for the null check (above), I think it will result in a final script that is unchanged.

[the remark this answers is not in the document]
gmcelhanon: After making all the changes identified in this review, I just restored the security database and ran the new SQL script against it. I then extracted the XML metadata and confirmed -- there were no changes to the final state of the security metadata (after making these adjustments to the XSLT stylesheet and the input TPDM metadata file).

Application/EdFi.Ods.Extensions.TPDM/Artifacts/MsSql/Data/Security/1010-TPDM-ResourceClaims.sql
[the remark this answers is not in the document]
blmeyers: This is a bug with the TPDM load process then, it shouldn't be loading these as part of bootstrap.  We can spin up another ticket to look into that.  Bootstrap should be as limited as possible.

PR #68, opened 2020-07-28
[ODS-4366] Adjust the suite number position in the nuget name

semalaiappan:


semalaiappan:
After Ed-Fi-TeamCity-Configs build merge , all  EdFi.Installer.SandboxAdmin ,EdFi.Installer.SwaggerUI ,EdFi.Installer.WebApi are started failing , so i need to update packageId to fix that issue

PR #100, opened 2020-09-01
[ODS-4516] Adjust for JSON Differences in Swagger Landing Page

tbmeador-dlp:


tbmeador-dlp:
fixed in ODS-4435 by modifying the JSON serialization to use camel case

PR #108, opened 2020-09-11
[ODS-4450] [ODS-4293] Verify that security headers are being stripped

tparandiev:


Application/EdFi.Ods.WebApi.NetCore/Program.cs
```diff
@@ -34,7 +34,16 @@ public static async Task Main(string[] args)
                             .AddEnvironmentVariables();
                     })
                 .UseServiceProviderFactory(new AutofacServiceProviderFactory())
-                .ConfigureWebHostDefaults(webBuilder => { webBuilder.UseStartup<Startup>(); }).Build();
+                .ConfigureWebHostDefaults(
```
jamessmckay, line 37: we should add this change to the startup for the SwaggerUI project too please

jamessmckay (changes requested):

Application/EdFi.Ods.WebApi.NetCore/Program.cs
tparandiev: okay Jim I will add it

jamessmckay (approved):

PR #137, opened 2020-10-09
[ODS-4449] Testing Reverse Proxy Support

blmeyers:
Adding launch profiles to all three projects to support local reverse proxy testing.


tbmeador-dlp (changes requested):
Needs a rebase and update to the launchsettings since the project rename went in.

blmeyers:
Rebased, Git picked up the rename so the launch settings adjusted.  Readme was already written assuming the rename so no changes there.  Let me know if anything else needs adjusting.

tbmeador-dlp (approved):

PR #147, opened 2020-10-16
[ODS-4572] Hashed secrets should be the default for non sandbox mode

jamessmckay:


logistics/scripts/modules/settings/settings-management.psm1
```diff
@@ -13,6 +13,7 @@ function Get-DefaultDevelopmentSettingsByProject {
             Urls              = "http://localhost:54746"
             ApiSettings       = @{
                 Engine = ""
+                DecryptSecrets = $true
```
vimayya, line 16: I would call it PlainTextSecret - This feature either stores hash or plaintext secrets, there is no encryption going on here.

jamessmckay: done

tnunnerydl (approved):

PR #150, opened 2020-10-21
[ODS-4578] Bulk Load Client Errors

jamessmckay:


logistics/scripts/modules/settings/settings-management.psm1
```diff
@@ -17,6 +17,7 @@ function Get-DefaultDevelopmentSettingsByProject {
             }
             ConnectionStrings = @{ }
             Plugin            = @{
+                Folder = "../../Plugin"
```
blmeyers, line 20: So what should the default Folder be for the appsettings.json?

blmeyers (changes requested):

logistics/scripts/modules/settings/settings-management.psm1
jamessmckay: good question, i need this to make the populated template to work. but I was unsure if I should change production version of the link in the appsettings.json.

blmeyers: This folder is moving anyways into WebApi, so this is fine for now.

blmeyers (approved):

blmeyers:
Issues with Context usage will be addressed in a future ticket.  Merging in so we can get templates building again.

PR #156, opened 2020-10-23
[ODS-3873] Use EdFi.Common package in ODS solution

semalaiappan:


blmeyers (changes requested):
As mentioned in the other PR, will need to bump code gen version here.

blmeyers (approved):

PR #185, opened 2020-11-13
[ODS-4673] Swagger UI Does not work with Docker

tnunnerydl:


jamessmckay (changes requested):
Should we remove the middleware?

jamessmckay (approved):

PR #194, opened 2020-12-04
[ODS-4594] Support for different virtual path with Reverse Proxy

tnunnerydl:


blmeyers (changes requested):

Application/EdFi.Ods.SandboxAdmin/Startup.cs
```diff
@@ -191,10 +192,19 @@ public void ConfigureContainer(ContainerBuilder builder)
...
-        public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerFactory loggerFactory)
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerFactory loggerFactory, IConfigurationRoot configuration)
         {
             loggerFactory.AddLog4Net();
 
+            var pathBase = configuration["PathBase"];
+
+            if (!string.IsNullOrEmpty(pathBase))
+            {
+                pathBase = pathBase.Replace("/", "");
```
blmeyers, line 203: Doing the slash handling this way is dangerous if they want to use a multipart path base such as "api/test".  This should only be normalizing trailing and leading slashes.

blmeyers (changes requested):

blmeyers (approved):

PR #198, opened 2020-12-08
[ODS-4486] Move packages to Azure Artifacts Update Kotlin Builds

gmcelhanon:
See also: https://github.com/Ed-Fi-Alliance/Ed-Fi-TeamCity-Configs/pull/31/files


tbmeador-dlp (changes requested):
It might be safe to remove this parameter from the build scripts but there could be other dependencies on the create-package.ps1 script. If this isn't required for the builds to work can we just close this PR?

gmcelhanon:
Leaving unchanged so as to not unnecessarily impact any existing usage of the scripts.

PR #200, opened 2020-12-10
[ODS-4692] RestAPI.Database package doesn't work with "OdsDatabaseTemplateName" as designed

raghuramakrishna:


tbmeador-dlp (changes requested):
To be consistent lets also add this settings to the WebApi settings with the default "minimal" https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/8e158e96d5fcc5d6b1f42beaae44490a542f9764/Application/EdFi.Ods.WebApi/appsettings.json

raghuramakrishna:
@tbmeador-dlp I have added OdsDatabaseTemplateName key in appsettings

tbmeador-dlp (changes requested):
Tested with it set to "populated" and it is still using the minimal template as it's backup source:
image
I think this is because the `Deployment.psm1` script has a parameter with a default value as "minimal". Can you test if removing the default allows you to restore from the populated template?
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/68f4ad6e2fa1ae823a461b426d0a3a0469cc254b/Scripts/NuGet/EdFi.RestApi.Databases/Deployment.psm1#L60

raghuramakrishna:
@tbmeador-dlp removed the default value. Please review.

tbmeador-dlp (approved):
Looks good!

PR #206, opened 2020-12-16
[ODS-4707] Support Configuration to Drive Reverse Proxy Support

raghuramakrishna:


jamessmckay (changes requested):
This implementation is not correct. What we need changed is in the startup class. of the application. In this case we need to wrap the registration of the proxy support with the check from the configuration. Secondly, then the AllForwarded headers being allowed needs to change to an logical or statement on the headers we are interested in.

jamessmckay (changes requested):
This looks good, however we need a pull request for the docker config being updated.

jamessmckay (changes requested):
Note, I just noticed the RequestExtensions shoud probably have the boolean value passed down from the controllers. The controllers should check for the key out of the config.

blmeyers (approved):

tnunnerydl (approved):

PR #214, opened 2021-01-13
[ODS-4735] Update TeamCity Builds to Use initdev

tbmeador-dlp:


Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -427,19 +422,117 @@ function Invoke-PesterTests {
...
+        }
+        New-Package @params | Out-Host
+    }
+}
+
+function New-WebPackage {
+    param(
+        [string] $ProjectPath,
+
+        [string] $PackageDefinitionFile = "$ProjectPath/bin/**/**/publish/$(Split-Path $ProjectPath -Leaf).nuspec",
```
blmeyers, line 491: Double ** here?  Does it not work as a 0 to many folder expansion for powershell?

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -427,19 +422,117 @@ function Invoke-PesterTests {
...
+        )
+
+        Write-Host -ForegroundColor Magenta "& dotnet $params"
+        & dotnet $params | Out-Host
+        Write-Host
+
+        $PackageDefinitionFile = (Get-ChildItem $PackageDefinitionFile)
+        if (-not [string]::IsNullOrWhiteSpace($PackageId)) {
+            [xml] $xml = Get-Content $PackageDefinitionFile
+            $xml.package.metadata.id = $PackageId
```
blmeyers, line 521: This seems suspect that we're modifying a source controlled file and changing it.  Why can't we pass the id along like you do above for the database package?  Do we need to update the checked in nuspec to have a variable (with a default here possibly) to allow better control of this?

build.teamcity.ps1
```diff
@@ -0,0 +1,72 @@
...
+
+# Package
+$params = @{
+    ProjectPath     = (Get-RepositoryResolvedPath (Get-ProjectTypes).SandboxAdmin)
+    PackageId       = Get-ValueOrDefault                  $teamcityParameters['odsapi.build.package.sandboxAdmin.id']
+    Version         = Get-ValueOrDefault                  $teamcityParameters['odsapi.build.package.sandboxAdmin.version']     '0.0.0'
+    Properties      = Get-ValueOrDefault (ConvertTo-Array $teamcityParameters['odsapi.build.package.sandboxAdmin.properties']) (Get-DefaultNuGetProperties)
+    OutputDirectory = Get-ValueOrDefault                  $teamcityParameters['odsapi.build.package.sandboxAdmin.output']      "$PSScriptRoot/packages"
+}
+New-WebPackage @params
```
blmeyers, line 43: This can be a separate follow up ticket, but packaging should be part of initdev and just not run by default, similar to tests.  This is a lot of logic and powershell someone would have to duplicate in order to make another build system function.

build.teamcity.ps1
```diff
@@ -0,0 +1,72 @@
...
+    NoRebuild     = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.noRebuild'])     $false
+    NoDeploy      = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.noDeploy'])      $false
+    RunPester     = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.runPester'])     $true
+    RunDotnetTest = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.runDotnetTest']) $true
+    RunPostman    = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.runPostman'])    $true
+    RunSmokeTest  = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.runSmokeTest'])  $true
+    UsePlugins    = Get-ValueOrDefault (ConvertTo-Boolean $teamcityParameters['odsapi.build.usePlugins'])    $false
+}
+
+Invoke-Task "Remove-EdFiDatabases" { Remove-EdFiDatabases -Force -Engine $params.Engine }
```
blmeyers, line 30: Should this be an option to run as part of initdev as well?  This might actually be nice to use locally...  I guess you could always just call it explicitly for now.

build.teamcity.ps1
```diff
@@ -0,0 +1,72 @@
...
+
+& "$PSScriptRoot/Initialize-PowershellForDevelopment.ps1"
+
+$teamcityParameters = Get-TeamCityParameters
+Write-Host
+Write-Host "$($teamcityParameters.Count) TeamCity parameters found."
+Write-Host
+
+# Build and Test
+$params = @{
```
blmeyers, line 16: Probably a separate ticket, but have we reached a point where a configuration object makes more sense than just continuing to add more switches?  Particularly if packaging ends up collapsing in.

logistics/scripts/modules/database/database-management.psm1
```diff
@@ -730,7 +730,10 @@ Function Remove-Database {
         # NOTE: We use DROP DATABASE rather than the SMO's .KillDatabase() function because the former can
         #       drop a database stuck in "Restoring" state, but the latter cannot
         $dropDatabase = @(
-            "ALTER DATABASE [$databaseName] SET SINGLE_USER WITH ROLLBACK IMMEDIATE"
+            "IF DATABASEPROPERTYEX('$databaseName', 'Status') != 'RESTORING'"
+            "BEGIN"
+                "ALTER DATABASE [$databaseName] SET SINGLE_USER WITH ROLLBACK IMMEDIATE"
+            "END"
```
blmeyers, line 736: You fixed it!

logistics/scripts/modules/settings/settings-teamcity.psm1
```diff
@@ -0,0 +1,23 @@
...
+
+    if (-not (Test-TeamCityVersion)) { return @{ } }
+
+    function Get-Parameters([string] $Path) { return Get-Content $Path -Raw -Encoding UTF8 | ConvertFrom-StringData }
+
+    $result = @{ }
+
+    $result = Merge-Hashtables $result, (Get-Parameters $env:TEAMCITY_BUILD_PROPERTIES_FILE)
+    $result = Merge-Hashtables $result, (Get-Parameters $result['teamcity.configuration.properties.file'])
+    $result = Merge-Hashtables $result, (Get-Parameters $result['teamcity.runner.properties.file'])
```
blmeyers, line 20: So we're pulling it from file system only, not using the injected parameters TeamCity shoves into Powershell?  Probably more consistent this way, just curious.

blmeyers (changes requested):

logistics/scripts/modules/settings/settings-teamcity.psm1
tbmeador-dlp: I think in order to inject you would have to create a build step with all the variables you want to pass. As far as I know it's not like octopus where they've exposed a powershell variable that you can access from any script.

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
tbmeador-dlp: Nope, there is just one too many wildcards.

build.teamcity.ps1
tbmeador-dlp: one of the reasons I put it here was so that all the TeamCity parameters would be exposed in this one file. I think it makes sense to move these if we go with a config object.

build.teamcity.ps1
tbmeador-dlp: yeah I think that would be useful at some point. I held off here because to might be a little more destructive than some may like.

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
tbmeador-dlp: I think in other cases we're passing the `packageId` to `dotnet pack`, which works well, but here I'm using nuget.exe and I don't believe it allows you to parameterize the `id`. `dotnet pack` has the issue where you can't package something that isn't a project, specifically it requires at least a .csproj file, so it won't work for the databases package. I went with nuget because that's how it currently works with the build server. The build server also has a feature that is string replacing the id which I've recreated here.

I definitely would like to use `dotnet pack` for all of these in the future, but I went with reusing some of the functionality we already had.

blmeyers: You can parameterize the id and pass it as a property, if you just put it as a variable in the nuspec, but it requires you to always pass the id in since it needs to be replaced.  Taking a quick look around, it looks like we've gotten away from that with most of the nuspecs, so we can defer on revisiting this if it matches for now.

blmeyers (changes requested):
Looks good and additional tickets are open, just needs to be caught up due to merge conflicts and then it should be good to go in.

blmeyers (approved):

PR #221, opened 2021-02-03
[ODS-4757] Remove explicit TLS 1.1/1.2 Settings

raghuramakrishna:


tbmeador-dlp (changes requested):
Let's go ahead and close this PR as it looks like this still may be needed for some environments.

PR #232, opened 2021-02-23
[ODS-4819] Update Landing page to show only major and minor version

raghuramakrishna:


jamessmckay (changes requested):
Version 5.1.0 remains, this should be for >= 5.2

tbmeador-dlp (changes requested):
This makes me think you're still fetching `v5.2.0` but then just changing the route in the html to `v5.2`. is that correct? This will work as long as we have a `v5.2.0` route but won't work when we push `v5.3` to staging.
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/1dc1f0494cc9123ab139afdf5323bc751ad7a101/LandingPage/index.js#L95

The logging appears to correctly display the routes you're actually hitting.
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/1dc1f0494cc9123ab139afdf5323bc751ad7a101/LandingPage/index.js#L103-L109

jamessmckay (approved):

PR #254, opened 2021-03-18
[ODS-4871] Update Installers/NugetHelper to load release promoted packages by default

semalaiappan:
$parameters = @{
    DbConnectionInfo = @{
        Engine="SqlServer"
        Server="localhost"
        UseIntegratedSecurity=$true
    }
    InstallType = "Sandbox"
    WebApiFeatures = @{
            BearerTokenTimeoutMinutes="BearerTokenTimeoutMinutes"                                  
            ExcludedExtensionSources="GrandBend"
			Features= @{
                "changeQueries"= @{
                    Name= "changeQueries"
                    IsEnabled= $true
                    }
				"OpenApiMetadata"= @{
                    Name= "OpenApiMetadata"
                    IsEnabled= $false
                    }
				"composites"= @{
                    Name= "composites"
                    IsEnabled= $false
                    }
				"profiles"= @{
                    Name= "profiles"
                    IsEnabled= $false
                    }
				"identityManagement"= @{
                    Name= "identityManagement"
                    IsEnabled= $false
                    }
				"extensions"= @{
                    Name= "extensions"
                    IsEnabled= $false
                    }
				"ownershipBasedAuthorization"= @{
                    Name= "ownershipBasedAuthorization"
                    IsEnabled= $false
                    }
				"uniqueIdValidation"= @{
                    Name= "uniqueIdValidation"
                    IsEnabled= $false
                    }
            }
	}     
}

### DIFF
--- a/Scripts/NuGet/EdFi.Installer.WebApi/Install-EdFiOdsWebApi.psm1
+++ b/Scripts/NuGet/EdFi.Installer.WebApi/Install-EdFiOdsWebApi.psm1
@@ -369,6 +369,7 @@ function Get-WebApiPackage {
             PackageVersion = $Config.PackageVersion
             ToolsPath = $Config.ToolsPath
             OutputDirectory = $Config.DownloadPath
+            PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi%40Release/nuget/v3/index.json"
         }
         $packageDir = Get-NuGetPackage @parameters
         Test-Error
--- a/Scripts/NuGet/EdFi.Installer.WebApi/Install-EdFiOdsWebApi.psm1
+++ b/Scripts/NuGet/EdFi.Installer.WebApi/Install-EdFiOdsWebApi.psm1
@@ -160,7 +160,11 @@ function Install-EdFiOdsWebApi {
         # NuGet package version. If not set, will retrieve the latest full release package.
         [string]
         $PackageVersion,
-        
+
+        # NuGet package source . default value is set for release package source for installer .
+        [string]
+        $PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi%40Release/nuget/v3/index.json",
+                
         # Path for storing installation tools, e.g. nuget.exe. Default: "./tools".
         [string]
         $ToolsPath = "$PSScriptRoot/tools",
@@ -275,6 +279,7 @@ function Install-EdFiOdsWebApi {
         WebApplicationPath = $WebApplicationPath
         PackageName = $PackageName
         PackageVersion = $PackageVersion
+        PackageSource = $PackageSource
         ToolsPath = $ToolsPath
         DownloadPath = $DownloadPath
         WebSitePath = $WebSitePath
@@ -369,7 +374,7 @@ function Get-WebApiPackage {
             PackageVersion = $Config.PackageVersion
             ToolsPath = $Config.ToolsPath
             OutputDirectory = $Config.DownloadPath
-            PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi%40Release/nuget/v3/index.json"
+            PackageSource = $Config.PackageSource
         }
         $packageDir = Get-NuGetPackage @parameters
         Test-Error
--- a/Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
+++ b/Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
@@ -56,6 +56,7 @@ function Invoke-DifferentPackageSource {
         PackageSource  = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json"
     }
     Install-EdFiOdsWebApi @p
+}
 
 function Invoke-SeparateConnectionInfo {
     $p = @{
--- a/Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
+++ b/Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
@@ -141,15 +141,39 @@ function Invoke-FeatureOverride {
         WebApiFeatures = @{
             BearerTokenTimeoutMinutes="BearerTokenTimeoutMinutes"                                   
             ExcludedExtensionSources="GrandBend"
-            FeatureIsEnabled=@{
-                changeQueries = $true
-                openApiMetadata = $false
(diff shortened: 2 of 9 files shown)


semalaiappan:
Please use  EdFi.Suite3.Installer.WebApi.5.2.51 Installer version for this testing . 

image

Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
```diff
@@ -164,6 +179,7 @@ function Invoke-UninstallDifferentWebSite {
 
 try {
     switch ($Scenario) {
+        "DifferentPackageSource" { Invoke-DifferentPackageSource }
```
tbmeador-dlp, line 182: need to update the default case to include this new scenario

Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
```diff
@@ -44,6 +44,21 @@ function Invoke-NonDefaultApplication {
...
+    $p = @{
+        ToolsPath = "../../../tools"
+        InstallType = "SharedInstance"
+        DbConnectionInfo = @{
+            Engine="SqlServer"
+            Server="localhost"
+            UseIntegratedSecurity=$true
+        }
+        PackageName = "EdFi.Suite3.Ods.WebApi"
+        PackageVersion = "5.2.14321"
```
tbmeador-dlp, line 57: As long as your new package is the most recent we shouldn't need the version here. Since the package name defaults to the value you've set here I don't see a reason to specify it explicitly.

tbmeador-dlp (changes requested):
Please make similar changes to the SwaggerUI and SandboxAdmin installers also.

semalaiappan:
Please use EdFi.Suite3.Installer.SwaggerUI.5.2.40 Installer version for this testing .
image

semalaiappan:
use  EdFi.Suite3.Installer.SandboxAdmin.5.2.60 for testing

semalaiappan:
> Please make similar changes to the SwaggerUI and SandboxAdmin installers also.

I updated it . Thanks

Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
semalaiappan: I updated it  .Thanks

Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
semalaiappan: I updated it  .Thanks

Scripts/NuGet/EdFi.Installer.WebApi/test.ps1
```diff
@@ -128,15 +141,39 @@ function Invoke-FeatureOverride {
...
+                    IsEnabled= $false
+                    }
+				"ownershipBasedAuthorization"= @{
+                    Name= "ownershipBasedAuthorization"
+                    IsEnabled= $true
+                    }
+				"uniqueIdValidation"= @{
+                    Name= "uniqueIdValidation"
+                    IsEnabled= $true
+                    }
```
tbmeador-dlp, line 176: I think this belongs in your other ticket

tbmeador-dlp (changes requested):

Scripts/NuGet/EdFi.Installer.WebApi/Install-EdFiOdsWebApi.psm1
```diff
@@ -160,7 +160,11 @@ function Install-EdFiOdsWebApi {
         # NuGet package version. If not set, will retrieve the latest full release package.
         [string]
         $PackageVersion,
-        
+
+        # NuGet package source . default value is set for release package source for installer .
+        [string]
+        $PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi%40Release/nuget/v3/index.json",
```
tbmeador-dlp, line 166: Passing in the local feed here fails when it tries to list the package version inside the `Get-NuGetPackage` function. The function assumes that only one package will be listed as a result of `nuget list`, but since it searches for a substring of package name it picks up both EdFi.Suite3.Ods.WebApi, and EdFi.Suite3.Ods.WebApi.Prerelease. This isn't a problem for our release feed because it only contains EdFi.Suite3.Ods.WebApi.

Nuget install will automatically install the latest version if no `version` parameter is provided, so the following code can likely be removed:
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/b2ac2cf0e831aa5443115fafce3f969ccf357e5c/logistics/scripts/modules/packaging/nuget-helper.psm1#L105-L115

Also I think we can remove the check to see if it is already downloaded since `nuget install` does this by default.
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/b2ac2cf0e831aa5443115fafce3f969ccf357e5c/logistics/scripts/modules/packaging/nuget-helper.psm1#L117-L123

then change the parameters to only pass in `PackageVersion` if it is not null (this is marked as an optional parameter):
```powershell
$parameters = @(
    "install", $PackageName,
    "-source", $PackageSource,
    "-outputDirectory", $OutputDirectory
)
if ($PackageVersion) {
    $parameters += "-version"
    $parameters += $PackageVersion
}
```
and finally modifying the returned value to resolve without a version using wildcards:
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/b2ac2cf0e831aa5443115fafce3f969ccf357e5c/logistics/scripts/modules/packaging/nuget-helper.psm1#L135

tbmeador-dlp (changes requested):

semalaiappan:
@tbmeador-dlp    I removed if (-not $PackageVersion) check and also if (Test-Path $downloadedPackagePath) .

semalaiappan:
I tested for local and release  feed also . Thanks

logistics/scripts/modules/packaging/nuget-helper.psm1
```diff
@@ -101,38 +101,20 @@ function Get-NuGetPackage {
     )
 
     $nuget = Install-NuGetCli $ToolsPath
```
tbmeador-dlp, line 103: we should update the version in the script to download the appCommon package with your changes `1.4.1170`
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/c9eb7787b3863766f8a3732133ac5f5fad8d4b3c/Scripts/NuGet/prep-installer-package.ps1#L18

tbmeador-dlp (changes requested):

tbmeador-dlp (approved):

CSR2017:
@vimayya 
Reviewed the changes. Thanks

PR #264, opened 2021-03-30
[ODS-4742] PostgreSQL Connection Pooling

jamessmckay:


logistics/scripts/modules/settings/settings-management.psm1
```diff
@@ -165,10 +165,10 @@ function Get-DefaultConnectionStringsByEngine {
...
-                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Master])   = "Host=localhost; Port=5432; Username=postgres; Database=postgres;"
+                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Ods])      = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_{0};pooling=true; Minimum Pool Size=10; Maximum Pool Size=50; client encoding=utf-8;"
+                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Admin])    = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_Admin;pooling=true; Minimum Pool Size=10; Maximum Pool Size=50; client encoding=utf-8;"
+                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Security]) = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_Security;pooling=true; Minimum Pool Size=10; Maximum Pool Size=50; client encoding=utf-8;"
+                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Master])   = "Host=localhost; Port=5432; Username=postgres; Database=postgres;pooling=false; client encoding=utf-8;"
```
stephenfuqua, line 171: Are these extra settings are ignored when PGBouncer isn't in use?

jamessmckay: Not sure I understand your question, but this is client side pooling. In the case of the master db, we only want 1 connection, and therefore it is not pooled.

logistics/scripts/modules/settings/settings-management.psm1
```diff
@@ -165,10 +165,10 @@ function Get-DefaultConnectionStringsByEngine {
         }
         PostgreSQL = @{
             ConnectionStrings = @{
-                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Ods])      = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_{0};"
-                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Admin])    = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_Admin;"
-                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Security]) = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_Security;"
-                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Master])   = "Host=localhost; Port=5432; Username=postgres; Database=postgres;"
+                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Ods])      = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_{0};pooling=true; Minimum Pool Size=10; Maximum Pool Size=50; client encoding=utf-8;"
```
blmeyers, line 168: All four of these should have a space between the ';' and the pooling keyword to match the connection string style.  Also as per https://www.npgsql.org/doc/connection-string-parameters.html, the parameters should probably be "Pooling" and "Client Encoding".

logistics/scripts/modules/settings/settings-management.psm1
```diff
@@ -165,10 +165,10 @@ function Get-DefaultConnectionStringsByEngine {
...
+                ((Get-ConnectionStringKeyByDatabaseTypes)[(Get-DatabaseTypes).Admin])    = "Host=localhost; Port=5432; Username=postgres; Database=EdFi_Admin;pooling=true; Minimum Pool Size=10; Maximum Pool Size=50; client encoding=utf-8;"
```
blmeyers, line 169: What are we solving by setting Client Encoding?  This is probably worth a look in another ticket, but we need to address this in way more places than just around pooling.  I'd recommend dropping Client Encoding and handling encodings through another ticket.

blmeyers (changes requested):

logistics/scripts/modules/settings/settings-management.psm1
jamessmckay: Note, the connection string is parsed using ignore case, culture invariant. However I made the change.

logistics/scripts/modules/settings/settings-management.psm1
jamessmckay: Done.

logistics/scripts/modules/settings/settings-management.psm1
jamessmckay: Note pooling is defaulted as true; however i added it to the connection string to make it easier to understand it is on.

blmeyers (approved):

PR #277, opened 2021-05-06
[ODS-4753] Deployment.psm1 Documentation Clarification

semalaiappan:


Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -48,10 +48,12 @@ function Initialize-DevelopmentEnvironment {
         Builds the ODS/API solution and deploys the necessary databases in order to setup a complete development environment.
     .parameter InstallType
         The type of deployment to install: 'Sandbox', 'SharedInstance', 'YearSpecific', or 'DistrictSpecific'.
+        InstallType variable is an string variable ,so you can use single quotes or double quotes both are supported format ,when you set this one value in octopus.
```
tbmeador-dlp, line 51: quotes aren't required here even in Octopus so I think we can remove this

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -48,10 +48,12 @@ function Initialize-DevelopmentEnvironment {
...
     .parameter OdsTokens
         A semicolon-separated string of tokens to use when creating Ods database instances.
         For a year specific deployment a valid value could be '2013;2014;2015;2016;2017'.
         For a district specific deployment a valid value could be '255901;255902'.
+        OdsTokens variable is an array of string variable ,so please use single quotes which is supported format ,when you set this values in octopus.
```
tbmeador-dlp, line 56: I think we should update this to say that it can be either:


a semi-colon separated string which requires quotes 
OR
an array of strings

This applies to both PowerShell and Octopus so I don't think it makes sense to mention Octopus specifically here.

tbmeador-dlp (changes requested):

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
semalaiappan: I removed it

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
semalaiappan: Applied new change .

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -49,7 +49,9 @@ function Initialize-DevelopmentEnvironment {
     .parameter InstallType
         The type of deployment to install: 'Sandbox', 'SharedInstance', 'YearSpecific', or 'DistrictSpecific'.
     .parameter OdsTokens
-        A semicolon-separated string of tokens to use when creating Ods database instances.
+        A semicolon-separated string which requires single quotes or an array of strings for OdsTokens to use when creating Ods database instances.
+        Example for Array of Strings in this format @("2018","2019","2020")
+		Example for semicolon-separated string which requires single quote	in this format '2013;2014;2015;2016;2017'
```
tbmeador-dlp, line 54:
```suggestion
        Example for semicolon-separated string which requires single quote in this format '2013;2014;2015;2016;2017'
```

tbmeador-dlp (changes requested):

tbmeador-dlp (approved):

PR #286, opened 2021-06-09
[ODS-4949] Modify Load Tools Script to Use dotnet CLI.

tnunnerydl:


logistics/scripts/modules/LoadTools.psm1
```diff
@@ -26,31 +26,43 @@ function Invoke-BuildLoadTools {
...
+     if (!(Test-Path -LiteralPath $buildLogFilePath -PathType Leaf)) {
+        Write-Warning ("Cannot find the build log file at '$buildLogFilePath', so unable to determine if build succeeded or not.")
+        return
     }
+
+    # Get if the build succeeded or not.
+    [bool] $buildFailed = (Select-String -Path $buildLogFilePath -Pattern "Build FAILED." -SimpleMatch -Quiet)
+
+    if ($buildFailed) {
+        Write-Host 'Ensure that the Visual Studio SDK is installed.'
```
tbmeador-dlp, line 61: can we remove this? it doesn't make sense here.
```suggestion
```

tbmeador-dlp (changes requested):
Looks good overall. Does it work if we remove the `Invoke-RestoreLoadToolsPackages` function? They should be restored automatically by the `dotnet build` in the `Invoke-BuildLoadTools` function.

tnunnerydl:
It worked after removing Invoke-RestoreLoadToolsPackages.

logistics/scripts/modules/LoadTools.psm1
```diff
@@ -11,46 +11,48 @@ Import-Module -Force -Scope Global (Get-RepositoryResolvedPath "logistics\script
...
 function Invoke-BuildLoadTools {
     param(
         [hashtable] $config
     )
 
     $buildConfiguration = $config.buildConfiguration
+    $solutionPath = $config.loadToolsSolution
+    $verbosity = "minimal"
+
+    if ((Get-DeploymentSettings).Engine -eq 'PostgreSQL') { $buildConfiguration = 'Npgsql' }
```
tbmeador-dlp, line 23: This line seems to be causing the template builds to fail. We don't have a `Npgsql` build configuration anymore so we should be able to remove it.
```suggestion
```

tbmeador-dlp (changes requested):

tbmeador-dlp (approved):

PR #323, opened 2021-09-15
[ODS-5058] EdOrg Strategies - Implement code changes to redirect Filters to use the new tuple table only

semalaiappan:


Application/EdFi.Ods.Features.OwnershipBasedAuthorization/NHibernate/OwnershipBasedAuthorizationNHibernateConfigurationActivity.cs
```diff
@@ -58,7 +58,8 @@ public void Execute(object sender, BindMappingEventArgs e)
         private static bool IsEdFiQueryMappingEvent(BindMappingEventArgs e)
         {
             return (e.Mapping.@namespace.Equals(Namespaces.Entities.NHibernate.QueryModels.BaseNamespace) ||
-                    e.Mapping.@namespace.Equals(Namespaces.Entities.NHibernate.QueryModels.Views))
+                    e.Mapping.@namespace.Equals(Namespaces.Entities.NHibernate.QueryModels.Views) ||
+                    e.Mapping.@namespace.Equals(Namespaces.Entities.NHibernate.QueryModels.Tables))
                    && e.Mapping.assembly.Equals(Namespaces.Standard.BaseNamespace);
```
blmeyers, line 63: This change (and the matching one on ChangeQueries) is a good indicator that this isn't really implementation specific code, and should be exposed in the core libraries.  Convert this boolean method into an extension method on BindMappingEventArgs, and then change there and ChangeQueries to consume it, so we can manage what are "Ed-Fi" query mapping events in one place.

blmeyers (changes requested):

blmeyers:
As overall feedback when testing this, we seem to be mostly handling all the main test cases now, but we still seem to have a bit of a mess when we get more complicated scenarios with multiple ed org types assigned.  The way the joins have changed now from each join representing one authorization segment option (through subtype specific views) to using a single join to the shared tuple table means that if there are multiple paths to authorize a single row, we end up getting that row back independently for each valid authorization path, breaking paging and counts.  This is similar to the known bug around students, but is way worse since it crosses across all education organization types. CC: @vimayya

vimayya:
> we end up getting that row back independently for each valid authorization path, breaking paging and counts. This is similar to the k

@blmeyers Do you have a suggestions for fixing it? cc @gmcelhanon

PR #383, opened 2022-01-31
ODS-5242 - Fix Invoke-DotnetTest

bptillman:
This fixes an issue found where when invoking dotnet test there were errors since with anything past NET5, a reference assembly is put in a ref folder, and then dotnet test was trying to execute against those items, as seen in the screenshot below. This was resolved by filtering out the ref folder.

Please not that other tests will still be failing on this build and will be handled by later tickets.

On feature-net6 branch currently:
image

On this branch now:
image

### DIFF
--- a/logistics/scripts/run-tests.ps1
+++ b/logistics/scripts/run-tests.ps1
@@ -5,7 +5,7 @@
 
 & "$PSScriptRoot\modules\load-path-resolver.ps1"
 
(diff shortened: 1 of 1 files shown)


semalaiappan:
@bptillman  This change looks good for Invoke-DotnetTest fix

AxelMarquez (approved):

PR #384, opened 2022-01-31
[ODS-5269] Upgrade Glendale and Northridge to v5.3

vimayya:
Test with v5.3 or with removal of 1330-UpdateVersionTo54.sql to avoid running into upgrade scenario.


bptillman (approved):
Changes look good. Both populated databases restored successfully and were able to be loaded.

PR #387, opened 2022-02-03
ODS-5260 Remove global.json files

semalaiappan:


bptillman (approved):
Confirmed no more global.json files exist in repo and everything is working like it was before the change.

PR #394, opened 2022-02-14
[ODS-5287] Sandbox Admin throws an exception during startup

AxelMarquez:
[UseHangfireServer got deprecated in favor of AddHangfireServer](https://www.hangfire.io/blog/2021/08/30/hangfire-1.7.25.html), however, there is already a call to _AddHangfireServer_ so the call to _UseHangfireServer_ can be safely removed.


bptillman (approved):
Changes look good and Sandbox is working for both Postgres and MSSQL

PR #406, opened 2022-03-11
ODS-5306 Caution note on the landing page

semalaiappan:
Use : CTRL+F5 : to refresh the page by clearing GG chrome cache. So that it apply new CSS file for this change


vimayya:
@semalaiappan the message here is different than the one in the swagger page. gist of the message is provided in the ticket.

bptillman (approved):

PR #408, opened 2022-03-23
[ODS-5305] Ed-Fi Demo Sandbox Refresh is not working

ericstockdale:


Application/EdFi.Ods.SandboxAdmin/Initialization/InitializationEngine.cs
```diff
@@ -169,6 +170,13 @@ public void RebuildSandboxes()
                     foreach (var sandbox in user.Value.Sandboxes.Where(x => x.Value.Refresh))
                     {
                         _log.Debug($"Resetting sandbox {sandbox.Key} for {clientProfile.Vendor.VendorName}");
+
+                        if (resetSandboxKeys.Contains(sandbox.Key))
+                        {
+                            continue;
```
bptillman, line 176: If we find a sandbox key here and skip resetting them by calling `continue`, then I think we should move the log statement above on line 172 to be after this. Since as it stands now its possible the log would say we were resetting a sandbox, but then not actually doing it.

ericstockdale: All set

bptillman (approved):

PR #422, opened 2022-05-03
[ODS-5363] Create a PowerShell function that returns whether it's running on Windows

jrogers-dlp:
This adds a new module, **cross-platform.psm1**, with a single function, **Get-IsWindows**,  to test if a machine is Windows without the automatic $IsWindows variable.

With this function, also adds:

- Validation to ensure Mono is installed on Unix-like systems
- Validation to ensure Postgres binaries are installed on Unix-like systems.

This has been tested on PS 5 on windows and PS 7.2 on Unix.

### DIFF
--- a/logistics/scripts/modules/database/postgres-database-management.psm1
+++ b/logistics/scripts/modules/database/postgres-database-management.psm1
@@ -24,10 +24,6 @@ function Test-PostgreSQLBinariesInstalled {
             return $true
         }
         catch {
-            Write-Host "ERROR:  Postgres client binaries are not installed on this Unix-like system." -ForegroundColor Red
-            Write-Host "To install them you can run:" -ForegroundColor Red
-            Write-Host "    Ubuntu: apt-get install postgresql-client" -ForegroundColor Red
-            Write-Host "    Alpine: apk add postgresql-client" -ForegroundColor Red
             return $false
         }
     }
@@ -67,7 +63,14 @@ function Install-PostgreSQLBinaries {
     $packagePath = Get-NuGetPackage @parameters
 
     if (-not (Test-PostgreSQLBinariesInstalled)) {
-        throw "Could not find PostgreSQL binaries in $script:toolsPath\$script:packageName\tools. "
+        if(Get-IsWindows){
+            throw "Could not find PostgreSQL binaries in $script:toolsPath\$script:packageName\tools. "
+        }else{
+            throw "ERROR:  Postgres client binaries are not installed on this Unix-like system.
+                    To install them you can run:
+                        Ubuntu: apt-get install postgresql-client
+                        Alpine: apk add postgresql-client"
+        }
     }
 }
 
--- a/logistics/scripts/modules/database/postgres-database-management.psm1
+++ b/logistics/scripts/modules/database/postgres-database-management.psm1
@@ -7,6 +7,7 @@ $ErrorActionPreference = "Stop"
 
 & "$PSScriptRoot\..\..\..\..\logistics\scripts\modules\load-path-resolver.ps1"
 Import-Module -Force -Scope Global (Get-RepositoryResolvedPath "logistics\scripts\modules\tasks\TaskHelper.psm1")
+Import-Module -Force -Scope Global (Get-RepositoryResolvedPath "logistics\scripts\modules\utility\cross-platform.psm1")
 Import-Module -Force -Scope Global (Get-RepositoryResolvedPath "logistics\scripts\modules\tools\ToolsHelper.psm1")
 Import-Module -Force -Scope Global (Get-RepositoryResolvedPath 'logistics\scripts\modules\packaging\nuget-helper.psm1')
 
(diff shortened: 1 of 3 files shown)


logistics/scripts/modules/database/postgres-database-management.psm1
```diff
@@ -7,6 +7,7 @@ $ErrorActionPreference = "Stop"
 
 & "$PSScriptRoot\..\..\..\..\logistics\scripts\modules\load-path-resolver.ps1"
```
stephenfuqua, line 8: Since you're already here, would it make sense to go ahead and change these to forward slashes now?

logistics/scripts/modules/database/postgres-database-management.psm1
```diff
@@ -17,6 +18,20 @@ $script:packageName = "PostgreSQL.Binaries"
...
+    if (!(Get-IsWindows)) {
+        try {
+            psql -V
+        }
+        catch {
+            Write-Host "ERROR:  Postgres client binaries are not installed on this Unix system." -ForegroundColor Red
+            Write-Host "To install them you can run:" -ForegroundColor Red
+            Write-Host "    Ubuntu: apt-get install postgresql-client" -ForegroundColor Red
+            Write-Host "    Alpine: apk add postgresql-client" -ForegroundColor Red
+            return
```
stephenfuqua, line 30: I agree with using `Write-Host` with foreground red, rather than `Write-Error`: the latter would actually raise an exception and cause execution to stop. 

However, need to return false here:

```suggestion
            return $false
```

logistics/scripts/modules/packaging/nuget-helper.psm1
```diff
@@ -3,6 +3,9 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
+& "$PSScriptRoot\..\..\..\..\logistics\scripts\modules\load-path-resolver.ps1"
+Import-Module -Force -Scope Global (Get-RepositoryResolvedPath "logistics\scripts\modules\utility\cross-platform.psm1")
```
stephenfuqua, line 7: Let's change those to forward slashes rather than propagating more backslashes...

logistics/scripts/modules/packaging/nuget-helper.psm1
```diff
@@ -25,6 +28,16 @@ function Install-NuGetCli {
         [string] $sourceNuGetExe = "https://dist.nuget.org/win-x86-commandline/v5.3.1/nuget.exe"
     )
 
+    if (!(Get-IsWindows)) {
+        try {
+            mono -V
+        }
+        catch {
+            Write-Host "ERROR:  Mono is not installed on this Unix system." -ForegroundColor Red
+            return
```
stephenfuqua, line 37: In this case, I think we want the process to stop: there is no point in continuing. Therefore I would change this to throw an exception. Before doing so, I would like to get a second opinion from someone on the Platform team when they review this PR. Also changing to "Unix-like" since almost no one actually uses Unix systems anymore 😉.

```suggestion
            throw "ERROR:  Mono is not installed on this Unix-like system."
```

logistics/scripts/modules/utility/cross-platform.psm1
```diff
@@ -0,0 +1,19 @@
...
+        Checks to see if the current machine is a Windows machine.
+    
+    .EXAMPLE
+        Get-IsWindows
+
+        returns $True
+    #>
+
+    if ($null -eq $IsWindows) {
+        $IsWindows = $true
```
stephenfuqua, line 13: This looks wrong. I know that it is actually right though. Please add comments explaining why this is right.

jrogers-dlp: add some comments explaining

logistics/scripts/modules/database/postgres-database-management.psm1
jrogers-dlp: nice catch, fixed this

logistics/scripts/modules/packaging/nuget-helper.psm1
jrogers-dlp: Ah yeah, my mistake, the throw just slipped my mind. fixed

logistics/scripts/modules/database/postgres-database-management.psm1
jrogers-dlp: I was actually set to start doing this on the next ticket, but then realized while testing on Unix that it actually didn't cause issue...do we still want to do it anyway?

logistics/scripts/modules/packaging/nuget-helper.psm1
jrogers-dlp: I realized while testing on Unix that this actually didn't cause issue...do we still want to do it anyway?

AxelMarquez (changes requested):
Just a few minor changes

logistics/scripts/modules/database/postgres-database-management.psm1
```diff
@@ -17,6 +18,20 @@ $script:packageName = "PostgreSQL.Binaries"
 $script:packageVersion = "12.2.0"
 
 function Test-PostgreSQLBinariesInstalled {
+    if (!(Get-IsWindows)) {
+        try {
+            psql -V
+        }
+        catch {
+            Write-Host "ERROR:  Postgres client binaries are not installed on this Unix system." -ForegroundColor Red
```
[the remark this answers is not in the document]
jrogers-dlp: changed to Unix-like

logistics/scripts/modules/database/postgres-database-management.psm1
```diff
@@ -17,6 +18,20 @@ $script:packageName = "PostgreSQL.Binaries"
 $script:packageVersion = "12.2.0"
 
 function Test-PostgreSQLBinariesInstalled {
+    if (!(Get-IsWindows)) {
+        try {
+            psql -V
```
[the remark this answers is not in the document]
jrogers-dlp: add parenthesis and `| out-null` to the check

Moved the error message to install-* instead

AxelMarquez (approved):

PR #425, opened 2022-05-05
[ODS-5402] - Add workflow for Sandbox Admin Installer

bptillman:
This differs from the previous workflows in two regards:
1. It runs on `windows-latest` instead of `ubuntu-latest` since we have not ported over the necessary Powershell scripts to be compatible with linux. 
2. There is only one workflow that is manually triggered, and not a manually triggered and pull request workflow. This matches how it existed on TeamCity.

### DIFF
--- a/Scripts/NuGet/EdFi.Installer.SandboxAdmin/EdFi.Installer.SandboxAdmin.nuspec
+++ b/Scripts/NuGet/EdFi.Installer.SandboxAdmin/EdFi.Installer.SandboxAdmin.nuspec
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package>
   <metadata>
-    <id>EdFi.Installer.SandboxAdmin</id>
+    <id>EdFi.Suite3.Installer.SandboxAdmin</id>
     <version>1.0.0</version>
     <title>Ed-Fi SwaggerUI Installer Scripts</title>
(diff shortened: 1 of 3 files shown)


Scripts/NuGet/EdFi.Installer.SandboxAdmin/EdFi.Installer.SandboxAdmin.nuspec
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package>
   <metadata>
-    <id>EdFi.Installer.SandboxAdmin</id>
+    <id>EdFi.Suite3.Installer.SandboxAdmin</id>
```
semalaiappan, line 4: Can you use file-content-replacer  here https://github.com/rshillington/file-content-replacer  instead of hard coding on this file . 

**/*.nuspec

Find what:  ?<=<id>)(.*?)(EdFi)(?=\b.*</id>)

Replace with %odsapi.package.suitenumber% wih Suite3

semalaiappan (changes requested):
use file-content-replacer here [rshillington/file-content-replacer](https://github.com/rshillington/file-content-replacer) instead of hard coding on this file Scripts/NuGet/EdFi.Installer.SandboxAdmin/EdFi.Installer.SandboxAdmin.nuspec

PR #429, opened 2022-05-16
[ODS-3536] Case sensitivity with "Changes" feature

ericstockdale:


bptillman (approved):
Confirmed that when running I can see it checking the Changes folder, whereas before it was not.

image

The builds failing on this are unrelated and broken on main as well and are being addressed in ODS-5434

PR #451, opened 2022-06-16
[ODS-5456] Security Metadata - Resource Claim for people has wrong parent resource claim when TPDM extension is applied

gmcelhanon:
Package version bump for TPDM extension templates for PostgreSQL.


semalaiappan (approved):
I verified that this latest change in Ed-Fi-ODS-Implementation\Plugin\EdFi.Suite3.Ods.Extensions.TPDM.Core.1.1.0.6.0.22\Artifacts\PgSql\Data\Security/1040-TPDM-Ensure-people-claim-has-no-parent-resource-claim  . So  I am approving this change
image

PR #454, opened 2022-06-22
ODS-5446 Code and Dependency Review Workflows (ODS/API)

semalaiappan:


.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,48 @@
...
+            $is_pull_request_branch = 'False'
+            $is_pull_request_branch = git ls-remote --heads origin ${{ env.CURRENT_BRANCH }} | Select-String -Pattern $patternName -SimpleMatch -Quiet
+            Write-Host "pull request branch is '$($is_pull_request_branch -eq $true)'" -fore GREEN
+            if ($is_pull_request_branch -eq $true) {
+              Write-Host "pull request branch " + ${{ env.CURRENT_BRANCH }} + "exists" -fore GREEN
+              git fetch origin ${{ env.CURRENT_BRANCH }}
+              git checkout ${{ env.CURRENT_BRANCH }}
+            }
+      - name: Dependency Review ("Dependabot on PR")
+        uses: actions/dependency-review-action@a9c83d3af6b9031e20feba03b904645bb23d1dab  # v1.0.2
```
AxelMarquez, line 48: Does this action also work for C#? I believe it only checks npm and ruby packages.
We might need to use ``dotnet list package --vulnerable`` instead.

.github/workflows/CodeQL Security Scan.yml
```diff
@@ -0,0 +1,97 @@
...
+    
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'csharp' ]
```
AxelMarquez, line 34: Why is matrix needed here? I believe we can simply set ``languages: 'csharp'`` on the CodeQL initialization below.

AxelMarquez (changes requested):
Can you rename the dependencies.yml to something more explanatory like ``Dependencies Security Scan``?

Also, can you fix the vulnerability in SandboxAdmin by specifying the header?

.github/workflows/dependencies.yml
semalaiappan: yes  .  It's not based on languages . https://github.com/actions/dependency-review-action

semalaiappan:
@AxelMarquez   I updated include the X-Frame-Options header for Ed-Fi-ODS-Implementation\Application\EdFi.Ods.SwaggerUI\web.config
Ed-Fi-ODS-Implementation\Application\EdFi.Ods.WebApi\web.config   and rebase with main of latest . Ready for final review now

AxelMarquez (approved):

PR #463, opened 2022-07-07
NO MERGE - code quality analysis spike

stephenfuqua:
In this spike, I have added two code analyzers to the "Entry Point" projects: Microsoft's [standard C# analyzer](https://docs.microsoft.com/en-us/visualstudio/code-quality/roslyn-analyzers-overview?view=vs-2022), and the [SonarSource analyzer](https://www.sonarlint.org/). The spike introduces an `editorconfig` file with substantial rule customization. This customization is based on the actual code practices in Analytics Middle Tier and here in these three projects updated projects.

Most of the "little" fixes were applied automatically by running `dotnet format` in the three projects. Some of the files only show a change at the top of the file - because a byte order mark was removed. I have confirmed that `IdentityContext` previously had a BOM and the format command removed it. Another small change is an extra line at the end of the file. This is a standard convention in the Unix-like world; its absence can impact the ability to open or display the file properly in some tools.


Application/EdFi.Ods.SandboxAdmin/EdFi.Ods.SandboxAdmin.csproj
```diff
@@ -25,13 +25,22 @@
...
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="6.0.3" />
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="6.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Log4Net.AspNetCore" Version="6.1.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="Npgsql" Version="6.0.3" />
     <PackageReference Include="NHibernate" Version="5.3.11" />
+    <PackageReference Include="SonarAnalyzer.CSharp" Version="8.41.0.50478">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
```
stephenfuqua, line 43: Here are the three NuGet packages that would need to be added in all projects, if we come to agreement about this code analysis process.

Application/EdFi.Ods.SandboxAdmin/Initialization/InitializationEngine.cs
```diff
@@ -163,7 +163,26 @@ public void RebuildSandboxes()
...
+                    if (resetSandboxKeys.Contains(sandbox.Key))
+                    {
+                        continue;
+                    }
+                */
+#pragma warning restore S125 // Sections of code should not be commented out
+
+#pragma warning disable S3267 // Loops should be simplified with "LINQ" expressions
                 foreach (var user in _users)
+#pragma warning restore S3267 // Loops should be simplified with "LINQ" expressions
```
stephenfuqua, line 185: Read above. I might actually like the code suggestion, if I had bothered to rewrite the Linq statement in the "proper" (preferred) way.

Application/EdFi.Ods.SandboxAdmin/Models/RegisterModel.cs
```diff
@@ -1,7 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
```
stephenfuqua, line 4: The code analyzer found a file without headers! Also, lots of unused imports.

Application/EdFi.Ods.SandboxAdmin/Models/Results/AdminActionResult.cs
```diff
@@ -23,10 +23,12 @@ public bool HasMessage
...
-        public string[] FailingFields
-        {
-            get { return _failingFields.ToArray(); }
-        }
+        // SF: I agree with this recommendation from SonarSource. However, I am nervous
+        // that this isn't being falgged as unused, yet changing to a method isn't
+        // breaking anything. Would need to test this thoroughly. Might prefer to
+        // make an exception and leave this alone. For more information, see 
+        // https://rules.sonarsource.com/csharp/RSPEC-2365
+        public string[] GetFailingFields() => _failingFields.ToArray();
```
stephenfuqua, line 31: Thoughts on this one?

Application/EdFi.Ods.SandboxAdmin/Services/PasswordService.cs
```diff
@@ -48,12 +48,10 @@ public async Task<PasswordResetResult> ValidateRequest(string userEmail, string
                 return PasswordResetResult.BadUsername;
             }
 
-            if (!await _identityProvider.VerifyUserEmailConfirmed(userEmail))
+            if (!await _identityProvider.VerifyUserEmailConfirmed(userEmail)
+                 && !await _identityProvider.VerifyUserPassword(userEmail, secret))
```
stephenfuqua, line 52: Code smell recommendation: unify the if statements. I agreed with that and applied this fix manually. This is not covered by `dotnet format`.

Application/EdFi.Ods.SandboxAdmin/Models/Results/AdminActionResult.cs
vimayya: used in javascript, so method may not be an option here

stephenfuqua: OK, I'll reverse that. Not that this will be merged - but I don't want anyone to copy and paste something inappropriate.

Application/EdFi.Ods.SandboxAdmin/Controllers/Api/ClientController.cs
```diff
@@ -92,17 +93,18 @@ public async Task<IActionResult> GetClients()
...
                     Trace.TraceError("{0}: {1}", currentException.Message, currentException.StackTrace);
-                    message += string.Format("{0}:{1}", counter, currentException.Message);
+                    message.AppendLine($"{counter}:{currentException.Message}");
                     counter++;
                     currentException = currentException.InnerException;
                 }
 
-                throw new Exception(string.Format("Unhandled exception with these messages: {0}", message));
+                // This is not an ideal exception type, but it is better than a plain Exception
+                throw new InvalidOperationException(string.Format("Unhandled exception with these messages: {0}", message));
```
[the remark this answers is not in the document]
gmcelhanon: This is where I disagree with the advice, and we've had this discussion previously, but to recap my opinion on this. I feel that there are only 2 times when a specifically typed exception is thrown in an _application_ (as opposed to a reusable library/framework):
- When there is an existing exception that matches perfectly as documented. `InvalidOperationException` is documented as "the exception that is thrown when a method call is invalid for the object's current state.". That doesn't fit here. There is nothing about the state of the controller that impacted the ability to invoke the `GetClients` method (controllers don't even have state, generally).
- When there is a need to attach definitive semantics to an exception so that it can be handled in a particular way elsewhere in the system (e.g. to be able to make a decision on which status code to return in the HTTP response). In most cases this is not needed (a 500 status is perfect) and the addition of new custom exception types for each occurrence is an onerous development and maintenance requirement, while an arbitrary usage of a _different_ type of the exception than `Exception` is pointless and equally lacking in semantic meaning.

Application/EdFi.Ods.SandboxAdmin/Controllers/Api/SandboxController.cs
```diff
@@ -10,7 +10,9 @@
 
 namespace EdFi.Ods.SandboxAdmin.Controllers.Api
 {
+#pragma warning disable S101 // Types should be named in PascalCase, however, uppercase is preferred for acronyms such as "DTO"
```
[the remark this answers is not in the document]
gmcelhanon: Personally, I dislike the "DTO" suffix entirely. I generally find it to be an unnecessary addition to the name.

Application/EdFi.Ods.SandboxAdmin/Extensions/HealthCheckServiceExtensions.cs
```diff
@@ -1,4 +1,4 @@
-﻿// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: Apache-2.0
```
gmcelhanon, line 1: Why are all these headers getting tweaked?

Application/EdFi.Ods.SandboxAdmin/Initialization/InitializationEngine.cs
gmcelhanon: I used to like the SQL-like syntax back in the day, but time has spoken and this style is rarely seen these days.

Application/EdFi.Ods.SandboxAdmin/Services/UserAccountManager.cs
```diff
@@ -72,9 +72,9 @@ public async Task<CreateLoginResult> Create(CreateLoginModel model)
                .AddFailingField(x => x.Email);
         }
 
-        public async Task<bool> Login(string userEmail, string password, bool isPersistent = false)
+        public async Task<bool> Login(string userName, string password, bool isPersistent)
```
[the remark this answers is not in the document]
gmcelhanon: Agree with the default arguments...

One semi-related question though... is the concept for that first argument best represented as "user name" or "username"?

Application/EdFi.Ods.SandboxAdmin/Controllers/Api/ClientController.cs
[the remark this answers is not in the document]
stephenfuqua: I actually agree with that position, though I had to jog my memory on those details by actually working through the example.

The only reason I would consider leaving this as a Suggestion (which, to be honest, might not ever be noticed) rather than completely disabling the advice is to help prompt people to _think_ about whether or not there is an existing appropriate exception.

Application/EdFi.Ods.SandboxAdmin/Services/UserAccountManager.cs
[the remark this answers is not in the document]
stephenfuqua: Good question, one I didn't put thought into because I wasn't going to change the signature in at the interface level in this POC.

Application/EdFi.Ods.SandboxAdmin/Extensions/HealthCheckServiceExtensions.cs
stephenfuqua: Because the original had a BOM, and `dotnet format` removed it.

PR #471, opened 2022-07-21
ODS-5476 Move ODS/API: Promote Packages build to Github Actions

semalaiappan:


.github/workflows/Promote Packages.yml
```diff
@@ -0,0 +1,50 @@
...
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+name: Promote Packages
+
+on:
+  workflow_dispatch:
+    inputs:
+      promote_packages_view:
+        description: ''select promote packages view for Azure Artifacts'
```
bptillman, line 12: There is an extra single quote at the start of this description that makes this yml file invalid, so it needs to be removed.

.github/workflows/Promote Packages.yml
```diff
@@ -0,0 +1,50 @@
...
+        required: true
+        default: 'Prerelease'
+        type: choice
+        options:
+        - Prerelease
+        - Release
+        - Local
```
bptillman, line 19: If we need to keep the Local option for testing then that is fine. But I tested in my own fork of the repo and only two options are required for the input to ensure its a drop down, as seen here 
image

bptillman (changes requested):
Just one required change to make here and a second optional one, depending on if it needs to stay around.

.github/workflows/Promote Packages.yml
semalaiappan: I removed it

.github/workflows/Promote Packages.yml
semalaiappan: I removed local

bptillman (approved):

PR #473, opened 2022-07-22
[ODS-5497] - Include ApiClientId in WebApi log statements

bptillman:
Now for any log entries that relate to a specific ApiClient, the ApiClientId will now get logged as part of each entry.
You can see it in the diff view, but the log entries will now look similar to this, where 2 is the ApiClientId on the first two entries since they relate to an ApiClient, while the third one does not have an ApiClientId since it is not related to an ApiClient:
```
2022-07-21 16:44:46,378 [.NET ThreadPool Worker,2] DEBUG AuthorizationBasisMetadataSelector - Authorization strategy 'NamespaceBased' selected for request against resource 'http://ed-fi.org/ods/identity/claims/ed-fi/assessment'.
2022-07-21 16:44:58,856 [.NET ThreadPool Worker,2] ERROR AssessmentsController - GetAllRequest
System.Exception: Exception of type 'System.Exception' was thrown.
2022-07-21 16:44:59,424 [.NET ThreadPool Worker,] DEBUG QuartzSchedulerThread - Batch acquisition of 0 triggers
```


Application/EdFi.Ods.WebApi/log4net.development.config
```diff
@@ -2,26 +2,26 @@
...
     <layout type="log4net.Layout.PatternLayout">
-      <conversionPattern value="%date [%thread] %-5level %logger - MESSAGE: %message%newline %exception" />
+      <conversionPattern value="%date [%thread,%property{ApiClientId}] %-5level %logger - MESSAGE: %message%newline %exception" />
     </layout>
   </appender>
   <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
     <threshold value="DEBUG" />
     <layout type="log4net.Layout.PatternLayout">
-      <conversionPattern value="%date [%thread] %-5level %logger{1} - %message%newline" />
+	  <conversionPattern value="%date [%thread,%property{ApiClientId}] %-5level %logger{1} - %message%newline" />
```
AxelMarquez, line 24: Please fix formatting

AxelMarquez (changes requested):
Now that we're at it can we also update the `log4net.config` for IntegrationTestHarness, CompositeSpecFlowTests, and WebApi.IntegrationTests so that they also log the ApiClientId?

Application/EdFi.Ods.WebApi/log4net.development.config
bptillman: This has been fixed.

bptillman:
> Now that we're at it can we also update the `log4net.config` for IntegrationTestHarness, CompositeSpecFlowTests, and WebApi.IntegrationTests so that they also log the ApiClientId?

Good catch. I have now added the ApiClientId to the pattern template

AxelMarquez (approved):

PR #487, opened 2022-08-08
ODS-5526 TPDM plugin is not included in WebAPI package and stage environment

semalaiappan:


bptillman (approved):
This change looks good. And thanks for explaining things to me over Slack. I've confirmed that before this change the Pester tests removed the plugins and after this change is applied the plugins remain.

PR #500, opened 2022-09-09
ODS-5571 Update launch settings for testharness to avoid hardcoded paths

semalaiappan:


bptillman (approved):
Changes look good to me.

PR #509, opened 2022-09-26
[ODS-5373] Support PS 7.2 and Unix - Script group 4

AxelMarquez:


Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -394,7 +360,7 @@ function Invoke-CodeGen {
         }
 
         $parameters = @(
-            "-r", $RepositoryRoot.ToLower(),
+            "-r", $RepositoryRoot,
```
AxelMarquez, line 363: This file isn't completely migrated yet, [ODS-5378](https://tracker.ed-fi.org/browse/ODS-5378) will complete its migration.
I've removed the `ToLower` so that we can at least run initdev in Linux and ease testing, note that `ToLower` isn't needed in Windows either.

DatabaseTemplate/Modules/get-populated-from-web.ps1
```diff
@@ -54,7 +54,7 @@ if ($fileName.EndsWith('.zip') -or $fileName.EndsWith('.7z')) {
     else {
         EnsureCommandIsAvailable "7z"
         $arguments = @("x", "$global:templateFolder/$fileName", "-o$global:templateDatabaseFolder")
-        7z @arguments
+        7z @arguments | Out-Null
```
AxelMarquez, line 57: This script doesn't belong to the current group, but while testing this group I found that 7z output was being returned to the caller, so I forwarded its output to null to be consistent with the behavior in Windows.

DatabaseTemplate/Modules/get-template-from-web.ps1
```diff
@@ -73,7 +73,7 @@ if ($isArchiveFile) {
     else {
         EnsureCommandIsAvailable "7z"
         $arguments = @("x", $archiveBackupFilePath, "-o$global:templateDatabaseFolder")
-        7z @arguments
+        7z @arguments | Out-Null
```
AxelMarquez, line 76: This script doesn't belong to the current group, but while testing this group I found that 7z output was being returned to the caller, so I forwarded its output to null to be consistent with the behavior in Windows.

logistics/scripts/modules/path-resolver.psm1
```diff
@@ -123,7 +123,7 @@ function Get-RepositoryNameFromPath ($path) {
     }
     #TODO:Enhance this logic so that it handles "C:/projects/Ed-Fi-Core" (no trailing slash) as an input
     foreach ($repoName in $repositoryNames) {
-        $repoRoot = "$(Get-RepositoryRoot $repoName)\"
+        $repoRoot = "$(Get-RepositoryRoot $repoName)$([IO.Path]::DirectorySeparatorChar)"
```
AxelMarquez, line 126: This is something that I should have caught when I reviewed this file in [ODS-5370](https://tracker.ed-fi.org/browse/ODS-5370); fixing it now.

Application/SolutionScripts/InitializeDevelopmentEnvironment.psm1
```diff
@@ -194,39 +193,6 @@ function Initialize-DevelopmentEnvironment {
     return $script:result | Format-Table
 }
 
-function Invoke-ConfigTransform {
```
AxelMarquez, line 197: Note that this file isn't completely migrated yet, [ODS-5378](https://tracker.ed-fi.org/browse/ODS-5378) will complete its migration.
I've just removed references to the deprecated scripts.

logistics/scripts/modules/database/postgres-database-management.psm1
```diff
@@ -5,42 +5,44 @@
...
-function Get-PSQLPath { return (Resolve-path (Join-Path (Get-PostgreSQLBinariesPath) 'psql.exe')).Path }
+function Get-PGDumpPath { return @(If (Get-IsWindows) { Join-Path $windowsPostgreSQLBinariesPath "pg_dump.exe" } Else { "pg_dump" }) }
 
-function Get-PGDumpPath { return (Resolve-path (Join-Path (Get-PostgreSQLBinariesPath) 'pg_dump.exe')).Path }
-
-function Get-PGRestorePath { return (Resolve-path (Join-Path (Get-PostgreSQLBinariesPath) 'pg_restore.exe')).Path }
+function Get-PGRestorePath { return @(If (Get-IsWindows) { Join-Path $windowsPostgreSQLBinariesPath "pg_restore.exe" } Else { "pg_restore" }) }
 
 function Install-PostgreSQLBinaries {
+    if(!(Get-IsWindows)) {
```
AxelMarquez, line 39: Moved this `if` up so that we fail fast in Linux.

logistics/scripts/modules/path-resolver.psm1
bptillman: Good catch

bptillman (changes requested):
I've finished my initial pass of the code itself and it mostly look good. I will start testing these in the VM now, but wanted to get these comments over to you sooner rather than later.

logistics/scripts/modules/postgres-binaries/build-postgres-binaries.psm1
```diff
@@ -5,15 +5,15 @@
...
-& "$PSScriptRoot\..\..\..\..\logistics\scripts\modules\load-path-resolver.ps1"
-Import-Module -Force -Scope Global (Get-RepositoryResolvedPath 'logistics\scripts\modules\tasks\TaskHelper.psm1')
-Import-Module -Force -Scope Global (Get-RepositoryResolvedPath 'logistics\scripts\modules\packaging\packaging.psm1')
+& "$PSScriptRoot/../../../../logistics/scripts/modules/load-path-resolver.ps1"
+Import-Module -Force -Scope Global (Get-RepositoryResolvedPath 'logistics/scripts/modules/tasks/TaskHelper.psm1')
+Import-Module -Force -Scope Global (Get-RepositoryResolvedPath 'logistics/scripts/modules/packaging/packaging.psm1')
 
 function Get-DefaultConfiguration([hashtable] $config = @{ }) {
     $config.binariesSourceUrl = 'http://get.enterprisedb.com/postgresql/'
     $config.binariesArchiveName = 'postgresql-{0}-1-windows-x64-binaries.zip'
```
bptillman, line 14: This is pulling the windows binaries specifically no matter which OS we run it on? I didnt see a 13.7 version version of the binaries for linux, so do we know if the windows ones will work for these purposes or not?

logistics/scripts/modules/postgres-binaries/build-postgres-binaries.psm1
```diff
@@ -91,15 +91,22 @@ function Get-PostgresBinaries([hashtable] $config = (Get-DefaultConfiguration))
 }
 
 function Expand-PostgresBinariesArchive([hashtable] $config = (Get-DefaultConfiguration)) {
-    if (-not (Get-InstalledModule | Where-Object -Property Name -eq 7Zip4Powershell)) {
+    if (Get-IsWindows -and -not Get-InstalledModule | Where-Object -Property Name -eq "7Zip4Powershell") {
```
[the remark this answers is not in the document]
AxelMarquez: Good catch, it's being transitively imported now but it's better to be explicit, I added the import.

logistics/scripts/modules/postgres-binaries/build-postgres-binaries.psm1
```diff
@@ -91,15 +91,22 @@ function Get-PostgresBinaries([hashtable] $config = (Get-DefaultConfiguration))
...
-
-    Get-7ZipInformation -ArchiveFileName $config.archivePath | Out-Host
-    Expand-7Zip -ArchiveFileName $config.archivePath -TargetPath $config.binariesTempDirectory
+    
+    if (Get-IsWindows){
+        Get-7ZipInformation -ArchiveFileName $config.archivePath | Out-Host
+        Expand-7Zip -ArchiveFileName $config.archivePath -TargetPath $config.binariesTempDirectory
+    }
+    else {
+        EnsureCommandIsAvailable "7z"
```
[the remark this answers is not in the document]
AxelMarquez: I'm not sure if it is a hard requirement, but I now print `7z l` output, which has the same information as `Get-7ZipInformation` plus more.

logistics/scripts/modules/postgres-binaries/build-postgres-binaries.psm1
AxelMarquez: This script builds `PostgreSQL.Binaries` package, which is only used in Windows. In Linux, we require the user to install PostgreSQL client binaries through their package managers (like apt); there's a validation in [postgres-database-management.psm1](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/81027a68246e1b978c05abf40a913a1b3664bfae/logistics/scripts/modules/database/postgres-database-management.psm1#L40) that ensures the binaries are installed.

bptillman (approved):

PR #513, opened 2022-09-28
[ODS-5556] Move EdFi.Ods.Minimal.Template Package build to Github Actions

ericstockdale:


configuration.packages.json
```diff
@@ -2,7 +2,7 @@
   "packages": {
     "EdFiMinimalTemplate": {
       "PackageName": "EdFi.Suite3.Ods.Minimal.Template",
-      "PackageVersion": "6.1.1",
+      "PackageVersion": "6.1.40",
```
bptillman, line 5: There look to be conflicts with this file that would prevent it from being merged to main currently. I also noticed how the initdev builds are failing which I think are related to this not being up to date as well. So I think what might need to happen is updating this from main, and then doing the usual build chain of updating these templates accorindgly.

DatabaseTemplate/Modules/create-database-template.psm1
```diff
@@ -512,15 +512,29 @@ function New-DatabaseTemplateNuspec {
     
     $nuspecPath = Join-Path $populatedTemplatePath "$packageNuspecName.nuspec"
 
+    $Id = '$id$'
```
bptillman, line 515: I wasnt able to get an idea for the reason behind this change from your commits, so could you explain to me what this change was for?

bptillman (changes requested):
This looks to be heading in the right direction. Just had one question on some code I was not sure of and one change that needs to be made related to a conflict in this branch and then I can re-review it.

DatabaseTemplate/Modules/create-database-template.psm1
ericstockdale: The nuspec file used to only allow for defaults that had to be replaced in the call to the nuget pack function.  Now that we are sharing the pack method with build.githubactions I placed the optional setting of these properties in the ps script while retaining the default behavior of nuspec file properties being set to placeholder values

bptillman (approved):

PR #520, opened 2022-10-07
ODS-5581 - Update step for checking out ods repo

bptillman:
We figured out there were a few situations that the way we were trying to determine which ODS branch to check out would not always work. Namely in these examples:
- If you created a branch for your work in the ODS repo but had no corresponding work in the Implementation repo, when we would try to check out the ODS repo in the workflow, it could not resolve the correct name, as seen [here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/actions/runs/3201463853/jobs/5229469442#step:7:1) .So then the ODS would end up staying on main instead of your feature branch
- Using HEAD_REF did not always resolve to the correct value, depending on how the workflow was triggered. In the case of it being triggered by a pull request, then HEAD_REF would correctly resolve to the feature branch we created that we opened the pull request for. But then in the context of being manually triggered, or being triggered from the ODS repo, it would resolve as an empty string, since HEAD_REF only works in the context of a pull request. Additionally, we could not switch to using REF_NAME since in the context of a pull request, it would resolve to something like refs/heads/548/merge, since that is the actual branch that triggered the workflow, not our feature branch. But otherwise REF_NAME would work for being triggered manually or from the ODS repo.

Additionally, in talking through it, we realized that when the ODS repo triggers a build in the Implementation repo, we would need to make sure Implementation was checked out to the right branch as well, so similar code was added to the "Is pull request in Ed-Fi-ODs-Implementation" step to see if an ODS branch was passed in the payload, and if so, check that out, otherwise use newer logic to determine the right branch to check out.

So the solution to this was to use a combination of two things:
- Short circuit all checks and first check for the ODS_BRANCH value being set or not, and if so, simply use it
- If not using the ODS_BRANCH, then check if REF_NAME resolved as something like refs/heads/548/merge and if so, use HEAD_REF instead as the branch we want to check out

To show this in action, outside of a workflow, here is some screenshots with a test script I wrote locally to mimic the changes. The test script looked like so:
```
$ref_name = 'ODS-5581b'
$head_ref = ''
$odsBranch = 'ODS-5581b'

Write-Host "OdsBranch: $odsBranch"
if($odsBranch -ne ''){
    Write-Host "Checking out ODS branch of $odsBranch"
    git fetch origin $odsBranch
    git checkout $odsBranch
} else {
    Write-Host "ref_name: $ref_name"
    $current_branch = $ref_name
    if ($current_branch -like "*/merge"){
    Write-Host "ref_name is PR, so using head_ref: $head_ref"
    $current_branch = $head_ref
    }
              
    Write-Host "Current Branch to try and checkout: $current_branch"
    $patternName = 'refs/heads/' +  $current_branch
    $branch_exists = 'False'
    $branch_exists = git ls-remote --heads origin $current_branch | Select-String -Pattern $patternName -SimpleMatch -Quiet
    if ($branch_exists -eq $true) {
    Write-Host "Current branch exists in Ed-Fi-ODS remote, so setting to $current_branch"
    git fetch origin $current_branch
    git checkout $current_branch
    } else {
        Write-Host "did not match on any results for changing ODS checkout branch"
    }
}
```

That example there has the three variables at the top set the way they would be if there was an ODS branch that triggered the Implementation repo that also had a matching branch name, which results in the following results where you can see I started out on main, and then successfully switched to ODS-5581b
image

Then if the variables are set like so, which would mimic this being triggered by a pull request on the implementation repo:
```
$ref_name = 'refs/heads/548/merge'
$head_ref = 'ODS-5581b'
$odsBranch = ''
```
When the script runs you can see I end up on the ODS-5581b branch:
image

Then if the variables are set like so, which would mimic manually triggering the implementation repo build, on a branch
```
$ref_name = 'ODS-5581b'
$head_ref = ''
$odsBranch = ''
```

You correctly end up on the corresponding ODS feature branch:
image

Then if the variables are set like so, which would mimic manually triggering the implementation repo build, on main
```
$ref_name = 'main'
$head_ref = ''
$odsBranch = ''
```

You correctly stay on main for ODS:
image

### DIFF
--- a/.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
+++ b/.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
@@ -25,7 +25,7 @@ env:
   VERSION_MINOR: "1"
   HEAD_REF:  ${{ GITHUB.HEAD_REF }}
   REF_NAME:  ${{ GITHUB.REF_NAME }}
-  ODS_CURRENT_BRANCH: ${{ github.event.client_payload.ref }}
+  ODS_CURRENT_BRANCH: ${{ github.event.client_payload.branch }}
   EDFI_SUITE_NUMBER: "Suite3"
 
 jobs:
--- a/.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
+++ b/.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
@@ -49,13 +49,33 @@ jobs:
       working-directory: ./Ed-Fi-ODS-Implementation/
       shell: pwsh
       run: |
-            $patternName = 'refs/heads/' +  '${{ env.CURRENT_BRANCH }}'
-            Write-Host "Pattern Name is " $patternName -fore GREEN
-            $is_pull_request_branch = 'False'
-            $is_pull_request_branch = git ls-remote --heads origin ${{ env.CURRENT_BRANCH }} | Select-String -Pattern $patternName -SimpleMatch -Quiet
-            if ($is_pull_request_branch -eq $true) {
-              git fetch origin ${{ env.CURRENT_BRANCH }}
-              git checkout ${{ env.CURRENT_BRANCH }}
+            $odsBranch = '${{ env.ODS_CURRENT_BRANCH }}'
+            Write-Host "OdsBranch: $odsBranch"
+            if($odsBranch -ne ''){
+              $patternName = "refs/heads/$odsBranch"
+              $does_corresponding_branch_exist = $false
+              $does_corresponding_branch_exist = git ls-remote --heads origin $odsBranch | Select-String -Pattern $patternName -SimpleMatch -Quiet
+              if ($does_corresponding_branch_exist -eq $true) {
+                Write-Host "Corresponding branch for $odsBranch exists in Implementation repo, so checking it out"
+                git fetch origin odsBranch
+                git checkout odsBranch
+              }
+            } else {
+              Write-Host "ref_name: ${{env.REF_NAME}}"
+              $current_branch = '${{env.REF_NAME}}'
+              if ($current_branch -like "*/merge"){
+                Write-Host "ref_name is PR, so using head_ref: ${{env.HEAD_REF}}"
+                $current_branch = '${{env.HEAD_REF}}'
+              }
+              $patternName = "refs/heads/$current_branch"
+              Write-Host "Pattern Name is " $patternName -fore GREEN
+              $is_pull_request_branch = $false
+              $is_pull_request_branch = git ls-remote --heads origin ${{ env.CURRENT_BRANCH }} | Select-String -Pattern $patternName -SimpleMatch -Quiet
+              if ($is_pull_request_branch -eq $true) {
+                Write-Host "Found Implementation branch with name ${{ env.CURRENT_BRANCH }} so checking it out"
+                git fetch origin ${{ env.CURRENT_BRANCH }}
+                git checkout ${{ env.CURRENT_BRANCH }}
+              }
             }
     - name: Checkout Ed-Fi-ODS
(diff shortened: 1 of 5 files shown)


.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
```diff
@@ -48,14 +49,38 @@ jobs:
...
-            $patternName = 'refs/heads/' +  '${{ env.CURRENT_BRANCH }}'
-            Write-Host "Pattern Name is " $patternName -fore GREEN
-            $is_pull_request_branch = 'False'
-            $is_pull_request_branch = git ls-remote --heads origin ${{ env.CURRENT_BRANCH }} | Select-String -Pattern $patternName -SimpleMatch -Quiet
-            if ($is_pull_request_branch -eq $true) {
-              git fetch origin ${{ env.CURRENT_BRANCH }}
-              git checkout ${{ env.CURRENT_BRANCH }}
-            }
+           $odsBranch = '${{ env.REPOSITORY_DISPATCH_BRANCH }}'
+           Write-Host "OdsBranch: $odsBranch"
```
stephenfuqua, line 53: https://github.com/PowerShell/PSScriptAnalyzer/blob/master/docs/Rules/AvoidUsingWriteHost.md

```suggestion
           Write-Output "OdsBranch: $odsBranch"
```

.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
```diff
@@ -48,14 +49,38 @@ jobs:
...
+              $branch_exists = $false
+              $branch_exists = git ls-remote --heads origin $current_branch | Select-String -Pattern $patternName -SimpleMatch -Quiet
+              if ($branch_exists -eq $true) {
+                Write-Host "Current branch exists, so setting to $current_branch"
+                git fetch origin $current_branch
+                git checkout $current_branch
+              } else {
+                Write-Host "did not match on any results for changing ODS checkout branch"
+              }
+           }
```
stephenfuqua, line 83: What do you think about moving most of this script into a dedicated PowerShell file? This way it could be analyzed with https://github.com/PowerShell/PSScriptAnalyzer when we eventually introduce it (surprise!). That also makes it easier to test the script locally without running in GitHub Actions.

I like to put such scripts into a `eng` directory at the root of the repo. That said, I realize there's an existing "logistics" root and it would be fine to put files there instead, in keeping with the repo's current organization.

This could be deferred to another ticket if needed. And I am making a suggestion, not a demand 😁 .

.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
```diff
@@ -48,14 +49,38 @@ jobs:
...
+                Write-Host "Corresponding branch for $odsBranch exists in Implementation repo, so checking it out"
+                git fetch origin $odsBranch
+                git checkout $odsBranch
+              } else {
+                WriteHost "Corresponding branch for $odsBranch does not exist in Implementation repo, so not changing branch checked out"
+              }
+           } else {
+              Write-Host "ref_name: ${{env.REF_NAME}}"
+              $current_branch = ${{env.REF_NAME}}
+              if ($current_branch -like "*/merge"){
```
AxelMarquez, line 68: We can use [github.event_name](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) to know if it comes from a PR (github.event_name == 'pull_request'); that way, the intent of this `if` would be much clearer.

.github/workflows/InitDev,Unit tests,Integration tests,Package.yml
bptillman: I agree! Especially since it is something I was already thinking about too.

bptillman: Okay this has been done.

AxelMarquez (approved):

PR #524, opened 2022-10-12
ODS-5566 Update Initdev Postgres Github Action workflows to use Postgres 13

semalaiappan:


.github/workflows/InitDev (Postgres), Tests using Sandbox, Year-Specific and Shared Instance.yml
```diff
@@ -11,11 +11,6 @@ on:
   push:
    branches: [main]
   workflow_dispatch:
-    inputs:
```
bptillman, line 14: Thanks for doing the cleanup here with the inputs the WriteHost as part of this work

bptillman (approved):
These changes look good to approve and merge.

PR #536, opened 2022-10-21
ODS-5610 Integration test harness appsettings.json file update to enable Profiles feature

semalaiappan:
…les feature


gmcelhanon:
I believe this is now obsolete based on ongoing work targeting the Dynamic Profiles epic. We can revisit if necessary.

PR #542, opened 2022-11-01
[ODS-5379] Support PS 7.2 and Unix - Script group 10

AxelMarquez:
Note that the next modified scripts don't belong to this group:
- create-database-template.psm1
- prep-package.ps1
- run-smoke-tests.ps1

... I modified them to call `GetExeExtension` instead of `Get-IsWindows` since it's a bit more readable.

### DIFF
--- a/Scripts/NuGet/EdFi.RestApi.Databases/PostDeploy.ps1
+++ b/Scripts/NuGet/EdFi.RestApi.Databases/PostDeploy.ps1
@@ -52,7 +52,5 @@ Add-Parameter $deploymentParams 'ExcludedExtensionSources' $ExcludedExtensionSou
 Add-Parameter $deploymentParams 'EnabledFeatureNames' $EnabledFeatureNames
 if ([Boolean]::TryParse((Add-Parameter $deploymentParams 'UsePlugins' $UsePlugins).UsePlugins, [ref]$UsePlugins)) { $deploymentParams.UsePlugins = $UsePlugins }
 
-If (-Not(Get-IsWindows)) {Add-Parameter $deploymentParams 'Engine' "PostgreSQL"} 
-
 Import-Module -Force -Scope Global "$PSScriptRoot/Deployment.psm1"
 Initialize-DeploymentEnvironment @deploymentParams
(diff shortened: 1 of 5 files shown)


Scripts/NuGet/EdFi.RestApi.Databases/PostDeploy.ps1
```diff
@@ -52,7 +52,5 @@ Add-Parameter $deploymentParams 'ExcludedExtensionSources' $ExcludedExtensionSou
 Add-Parameter $deploymentParams 'EnabledFeatureNames' $EnabledFeatureNames
 if ([Boolean]::TryParse((Add-Parameter $deploymentParams 'UsePlugins' $UsePlugins).UsePlugins, [ref]$UsePlugins)) { $deploymentParams.UsePlugins = $UsePlugins }
 
-If (-Not(Get-IsWindows)) {Add-Parameter $deploymentParams 'Engine' "PostgreSQL"}
```
AxelMarquez, line 55: This is breaking Octopus deployments since `cross-platform.psm1` is not imported, so `Get-IsWindows` isn't defined.
I'm reverting this change since we only use SQL Server in our Octopus deployments; consider adding an `Engine` parameter instead if Postgres has to be supported.

stephenfuqua:
> `create-database-package.ps1` is being left alone, since the only places it is referenced are the EdFi.Database.Security and EdFi.Database.Admin builds in TeamCity/Github, and the builds in Github will remain on `windows-latest` due to the requirement of needing the sql binaries for their database packages to be created.

Is there a ticket to look into changing that in the future? Surely there is an alternate method to get what we need.

bptillman:
> > `create-database-package.ps1` is being left alone, since the only places it is referenced are the EdFi.Database.Security and EdFi.Database.Admin builds in TeamCity/Github, and the builds in Github will remain on `windows-latest` due to the requirement of needing the sql binaries for their database packages to be created.
> 
> Is there a ticket to look into changing that in the future? Surely there is an alternate method to get what we need.

Yes @vimayya and I were just talking about how there is a sqlpackage available now in linux, so when we look at moving the relevant workflows from Windows to Ubuntu, we will see if using that will work so that the workflow does not have to remain on Windows.

bptillman (approved):

PR #545, opened 2022-11-08
ODS-5380 Support PS 7.2 and Unix - Script group 11

semalaiappan:


logistics/scripts/Invoke-PostmanIntegrationTests.ps1
```diff
@@ -14,24 +14,32 @@
...
-        npm install -g newman-reporter-teamcity@0.1.12
-        npm install -g newman-reporter-junitfull
-        newman --version
+        if (Get-IsWindows){
+            npm install -g newman@5.2.2 --verbose
+            npm install -g newman-reporter-teamcity@0.1.12
+            npm install -g newman-reporter-junitfull
+            newman --version
+        } else {
+            sudo npm install -g newman@5.2.2 --verbose
```
bptillman, line 38: Calling sudo here works in the context of Github Actions since it runs a [passwordless sudo](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#administrative-privileges) , but if someone were trying to run Postman tests in their Linux environment, or possibly on a Mac, this would mean that the build script would pause at this point and ask for their password. Instead, for the non-Windows side I think we should utilize the `EnsureCommandIsAvailable` to check if newman is installed already (similar to what we do for mono). Additionally, since the teamcity reporter is only needed for TeamCity and the junit one is only needed for Github Actions, we should break up this install to be specific to each environment.

Something in line with this, but possibly not exactly like this (where a Test-GithubActions method is added that checks for us being in github):
```
function Install-Newman {
    try {
        if (Get-IsWindows){
            npm install -g newman@5.2.2 --verbose
        } else {
            EnsureCommandIsAvailable "newman"
        }
        If(Test-TeamCityVersion){
            npm install -g newman-reporter-teamcity@0.1.12
        }
        If(Test-GithubActions){
            if(Get-IsWindows){
                npm install -g newman-reporter-junitfull
            }else {
                sudo npm install -g newman-reporter-junitfull
            }
        }
    }
    catch {
        Write-Host $_ -ForegroundColor Red
        Clear-Error
    }
}
```

bptillman (changes requested):

logistics/scripts/Invoke-PostmanIntegrationTests.ps1
bptillman: Another more verbose version, with some TODO's could look something like this
```
function Install-Newman {
  try {
    $packages = @{
      newman         = @{
        name            = "newman"
        requiredVersion = "5.2.2"
      }
      newmanTeamcity = @{
        name            = "newman-reporter-teamcity"
        requiredVersion = "0.1.12"
      }
      newmanJunit    = @{
        name            = "newman-reporter-junitfull"
        requiredVersion = "" # empty means any version (latest)
      }
    }
    $teamcityPackages = @($packages.newman, $packages.newmanTeamcity, $packages.newmanJunit)
    $githubActionsPackages = @($packages.newman, $packages.newmanJunit)
    $developerPackages = @($packages.newman)
    If (Test-TeamCityVersion) {
      Install-NpmPackages $teamcityPackages
    }
    elseIf (Test-GithubActions) {
      Install-NpmPackages $githubActionsPackages
    }
    else {
      Install-NpmPackages $developerPackages
    }
  }
  catch {
    Write-Host $_ -ForegroundColor Red
    Clear-Error
    # todo, should we remove this try and let the exception bubble so that it stops the build/script?
  }
}

function Install-NpmPackages {
  param(
    [Hashtable[]] $packages
  )
  foreach ($package in $packages) {
    if (Get-IsWindows) {
      & "npm install -g $(package.name)@$(package.name) --verbose"
    }
    else {
      # todo, check if the package is already installed using "npm ls -g", throw if not 
    }
  }
}
```

semalaiappan: As described , I am implemented to handle three different flows .

logistics/scripts/Invoke-PostmanIntegrationTests.ps1
```diff
@@ -14,33 +14,78 @@
...
+
+function Install-NpmPackages {
+  param(
+    [Hashtable[]] $packages
+  )
+  foreach ($package in $packages) {
+    $does_corresponding_package_exist = npm list -g --depth=0 |Out-String -Stream | Select-String -Pattern $package.name -SimpleMatch -Quiet
+    if (!$does_corresponding_package_exist -eq $true) {
+      if (Get-IsWindows) {
+        & "npm install -g '$($package.name)@$($package.requiredVersion)' --verbose"
```
AxelMarquez, line 78:
```suggestion
        npm install -g "$($package.name)@$($package.requiredVersion)" --verbose
```

bptillman (approved):

PR #557, opened 2022-11-30
[ODS-5409] Switch Initdev workflows from windows-latest to ubuntu-latest

AxelMarquez:


.github/workflows/InitDev (Postgres), Tests using Sandbox, Year-Specific and Shared Instance.yml
```diff
@@ -25,19 +25,32 @@ env:
   REPOSITORY_DISPATCH_BRANCH: ${{ github.event.client_payload.branch }}
   EDFI_SUITE_NUMBER: "Suite3"
   msbuild_buildConfiguration: "release"
+  LANG: "en_US.utf8"
```
AxelMarquez, line 28: There are a couple of Postman tests that do asserts on date strings, so dates have to be serialized using "en_US.utf8" format (instead of the default ""C.utf8"").

semalaiappan (approved):

PR #584, opened 2023-01-05
[ODS-5675] Include basic Request-Response details in the logs

simpat-jesus:


vimayya:
@simpat-jesus Let the default setup be at the ERROR level, you can close this PR.

simpat-jesus:
I will revert the log level to be ERROR; we need to keep the pattern layout change 
image

simpat-jesus:
Logging only ERROR messages by default
image

semalaiappan (approved):
I tested with both Error and Info level and both are working fine

PR #592, opened 2023-01-19
[ODS-5686] Minimal template should include only include descriptors with ed-fi namespace

ericstockdale:


semalaiappan (changes requested):
Please use Rebuild Database Templates workflow to update packages version in ODS implementation .   Pre-requisites is you need to have ODS-5686 branch in ODS , ODS implementation and ODS Extension repo ,then run  Rebuild Database Templates workflow  from ODS-5686 branch . It will take 1 hour to update all packages automatically for you .  Note  Please clean up the existing manual package version what you have in ODS implementation repo of ODS-5686 branch.  All 3 branch needs to pushed to remote to execute  Rebuild Database Templates workflow .

PR #608, opened 2023-02-02
[ODS-5718] Reorganize Ed-Fi Standard and Extension Artifacts using version-based folder structure

simpat-jesus:


mjaksn:
It looks like a trailing comma is causing an issue in Application\SolutionScripts\InitializeDevelopmentEnvironment.psm1

simpat-jesus:
Fixed the trailing comma issue

PR #609, opened 2023-02-08
Ed org updates

jonHickam:
Here are the updates to the district names and the population of the status charter for LEAs and schools.


stephenfuqua:
This PR was opened in the wrong repository.

PR #619, opened 2023-02-23
[ODS-5732] Trgr EdFi.Ods.DB.Templates workflow in ODS repo removed

semalaiappan:
Any code Changes to scripts in https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/tree/main/DatabaseTemplate/Modules possibly need template run and definitely handled by  manual run of [Rebuild Database Templates.yml](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/main/.github/workflows/Rebuild%20Database%20Templates.yml)  .so removing Trgr EdFi.Ods.DB.Templates workflow in ODS repo workflow


mjaksn (approved):
I agree with the reasoning given for removing this workflow from Ed-Fi-ODS-Implementation

PR #624, opened 2023-03-08
[ODS-5742] Nuget package caching enabled in Github Actions builds of Ed-Fi-ODS-Implementation repo

semalaiappan:
In general, it's not recommended to share the NuGet package cache across operating systems. NuGet packages may include platform-specific binaries or dependencies, and they may not work correctly on different operating systems.

Note that if you use the same cache key for multiple operating systems, you should make sure that the NuGet packages are compatible with all operating systems. Some packages may include platform-specific binaries or dependencies, and they may not work correctly on different operating systems. In that case, you should use separate cache keys for each operating system to ensure that the packages are only used on compatible systems. That's the reason we have 2 caches in this ODS Implementation Repo

### DIFF
--- a/.github/workflows/Cleanup Caches by a branch.yml
+++ b/.github/workflows/Cleanup Caches by a branch.yml
@@ -1,0 +1,46 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+name: Cleanup Caches by a branch
+on:
+  pull_request:
+    branches: [main, 'ODS-*']
+    types: [ closed ]
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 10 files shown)


mjaksn:
> 



> In general, it's not recommended to share the NuGet package cache across operating systems. NuGet packages may include platform-specific binaries or dependencies, and they may not work correctly on different operating systems.
> 
> Note that if you use the same cache key for multiple operating systems, you should make sure that the NuGet packages are compatible with all operating systems. Some packages may include platform-specific binaries or dependencies, and they may not work correctly on different operating systems. In that case, you should use separate cache keys for each operating system to ensure that the packages are only used on compatible systems. That's the reason we have 2 caches in this ODS Implementation Repo

Starting with v3.2.3 of actions/cache, cross-os cache restoration was introduced as an opt-in feature and will only apply if enableCrossOsArchive input is passed as true `https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cross-os-cache`

mjaksn:
@semalaiappan , can you rebase onto main? That should fix the checks. Then once they pass, I'll approve and merge

mjaksn (approved):

PR #626, opened 2023-03-10
[ODS-5764] Profile usage is not enforced (v5.1)

gmcelhanon:
Enabled Profiles feature on integration test harness.


semalaiappan:
All Unit Tests and Integration Tests are green except "POST Request contains a content type header using a profile that does not include the targeted resource"

semalaiappan (approved):

PR #642, opened 2023-04-01
[ODS-5766] Add the Standard version to the minimal and populated templates

simpat-jesus:
To test the database packages are loaded correctly, run initdev without the tpdm plugin, this should load minimal and populated templates for the specified database engine

image


semalaiappan (changes requested):

semalaiappan (approved):
image
Working fine

PR #647, opened 2023-04-11
[ODS-5768]  Add standard version to package name in packages generated on InitDev,Unit tests,Integration tests,Package workflow

semalaiappan:


semalaiappan:
image

simpat-jesus (changes requested):
Looks good, just update the action name to follow what we have been doing in previous builds, ex. build (Standard 4.0.0)

simpat-jesus (approved):

PR #661, opened 2023-04-25
[ODS-5812] Replace System.Data.SqlClient with Microsoft.Data.SqlClient in solution

simpat-adam:


semalaiappan (changes requested):
https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/f673705188edccae27160badd32e0399e05dedef/configuration.packages.json#L70 you need to update the 
EdFi.Suite3.Db.Deploy.4.0.23 package version  under 
"PackageName": "EdFi.Suite3.Db.Deploy" in Configuration Json file

DatabaseTemplate/Modules/create-database-template.psm1
```diff
@@ -477,7 +477,7 @@ function Backup-DatabaseTemplate {
 
     if ($config.engine -eq 'SQLServer') {
         $params = @{
-            csb               = Convert-CommonDbCSBtoSqlCSB $csb
+            csb               = $csb
```
semalaiappan, line 480: image
This file is used various database template packages ,so you need to run the database template workflow to update the package version for all database template packages

semalaiappan (changes requested):
you need to run the database template workflow to update the package version for all database template packages since there is change in create-database-template.psm1

semalaiappan (approved):

PR #707, opened 2023-06-27
[ODS-5912] Bring in Parent to Contact change 

semalaiappan:
https://intedfitools1.msdf.org/buildConfiguration/EdFi_DataStandardTeam_DataModelBuilds_Extensions_MetaEdJsExtensionArtifactsSample/198086?buildTab=log&focusLine=283&linesState=167&logView=flowAware

error:     Association additions 'StudentParentAssociation' does not match any declared Association or Association Subclass in namespace EdFi.  D:\BuildAgent\work\7fb716618dd0e578\MetaEdExtensionSource\Association\StudentParentAssociationExtension.metaed (6:13)
  error:     Association property 'StudentParentAssociation' does not match any declared Association or Association Subclass in namespace EdFi.  D:\BuildAgent\work\7fb716618dd0e578\MetaEdExtensionSource\Association\StudentGraduationPlanAssociation.metaed (63:22)
  error:     'common extension' is invalid for property Address on Domain Entity Extension Parent. 'common extension' is only valid for referencing common properties with the same declaration as in the base entity.  D:\BuildAgent\work\7fb716618dd0e578\MetaEdExtensionSource\DomainEntity\ParentExtension.metaed (63:5)
  error:     Domain Entity additions 'Parent' does not match any declared Domain Entity or Domain Entity Subclass in namespace EdFi.  D:\BuildAgent\work\7fb716618dd0e578\MetaEdExtensionSource\DomainEntity\ParentExtension.metaed (6:15)

image
image


Extensions/Sample/SampleMetaEd/Association/StudentContactAssociationExtension.metaed
```diff
@@ -3,34 +3,34 @@
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
 
-Association EdFi.StudentParentAssociation additions
+Association EdFi.StudentContactAssociation additions
```
gmcelhanon, line 6: Since these MetaEd files aren't segmented by standard version, how can we possibly support the extension to run against any version of the Ed-Fi Standard? It seems to me that this change would break our ability to generate artifacts for use with the v4.0 and previous standards.

gmcelhanon (commented):
I just have a question about flexible version support with only one copy of these sample extensions in the repository.

Extensions/Sample/SampleMetaEd/Association/StudentContactAssociationExtension.metaed
semalaiappan: Yes .  But we are not generate artifacts for previous standard version (4.0.0 )in teamcity  for any extension.   Always we will have latest version (5.0.0)

semalaiappan: Do we still need to maintain MetaEd Files for each standard version in this repo , not sure whether required or not ?

gmcelhanon: This is probably something we'll need to discuss (revisit) with @vimayya when she is back.

If we were to change something about the technical implementation of the API in the future (e.g. perhaps we rename the base API controller), we would (theoretically) need to be able to regenerate the extensions artifacts to run under the v8 API with the v4.x of Ed-Fi Standard (i.e. this is the "flexible" data standard support we're adding in v7 API). With these changes, we would not be able to do that because the change in the model is very much a breaking change.

That said, this doesn't block us from merging this now and coming back to it.

gmcelhanon (approved):

Extensions/Sample/SampleMetaEd/Association/StudentContactAssociationExtension.metaed
semalaiappan: I updated in latest commit having copy of  4.0.0 MetaEd files for both extension under Ed-Fi-ODS-Implementation\Extensions\4.0.0\  .So that we can use it later if required since 4.0.0 also supported in 7.0.0 release

gmcelhanon (approved):

PR #759, opened 2023-08-04
[ODS-5898] Docker for Sandbox Admin

stephenfuqua:
This pull request adds Docker support for the Sandbox Admin tool, including local testing capabilities. In addition, it:

* loads the minimal template in the ODS host along with populated for sandbox mode;
* refactors a number of environment variables for better standardization;
* removes NGiNX as a gateway, to reinforce that docker compose files are for local testing only;
* fixes a bug in the semantic version sorting of the get-version script.


Docker/.gitignore
stephenfuqua, line 1: Renamed the based directory from `docker` to `Docker` to be consistent with the other upper-case directories in the repo.

Docker/README.md
stephenfuqua, line 1: This file is not new, but sadly GitHub thinks it is. It is lightly edited to reflect the changes made in this PR.

stephenfuqua:
The build failures do not appear to have anything to do with my code changes in this PR.

Docker/bootstrap.ps1
```diff
@@ -9,33 +9,34 @@
...
+    $Hub
 )
 
+$ErrorActionPreference = "Stop"
+
 if (-not (Test-Path ./gateway/ssl/server.crt)) {
     Push-Location gateway
     ./generate-cert.ps1
     Pop-Location
 }
```
mjaksn, line 22: This code block must be removed since the gateway container is no longer included.

mjaksn (changes requested):

Docker/get-versions.ps1
```diff
@@ -57,12 +65,17 @@ function Invoke-SemanticSort {
...
+    $major = @{label="major";expression={[int]$_.Split(".")[0]}}
+    $minor = @{label="minor";expression={[int]$_.Split(".")[1]}}
+    $patch = @{label="patch";expression={[int]$_.Split(".")[2]}}
+
     $Versions `
-        | Select-Object {$_.Split(".")} `
-        | Sort-Object {$_.'$_.Split(".")'[0], $_.'$_.Split(".")'[1], $_.'$_.Split(".")'[2]} -Descending `
-        | ForEach-Object { $_.'$_.Split(".")' -Join "." }
+        | Select-Object $major, $minor, $patch `
+        | Sort-Object major, minor, patch -Descending
```
mjaksn, line 74: A `` ` `` must be added at the end of this line to avoid a syntax error.

mjaksn (changes requested):

Docker/get-versions.ps1
stephenfuqua: That's strange - it sure looks like a syntax error, yet I've run this many times without a problem, on my localhost and on a VM. I'm running PowerShell 7; maybe it accepts the pipe without needing a backtick to continue on the next line.

mjaksn (approved):

PR #762, opened 2023-08-08
[ODS-5931] Add branch name to dispatched actions

mjaksn:


mjaksn:
I tested this using my own forks of the ODS and Implementation repositories.

simpat-jesus (approved):

PR #765, opened 2023-08-09
[ODS-5927] Database template builds should error on XML validation failures

mjaksn:
These changes come from a snippet shared by @gmcelhanon

### DIFF
--- a/logistics/scripts/modules/utility/xml-validation.psm1
+++ b/logistics/scripts/modules/utility/xml-validation.psm1
@@ -77,7 +77,7 @@ function Invoke-XmlValidation {
     [System.Xml.Schema.XmlSchemaValidationFlags]::ReportValidationWarnings
     $xmlReaderSettings.Schemas.XmlResolver = New-Object System.Xml.XmlUrlResolver
 
-    ($script:validationErrorCount) = 0
+    $script:validationErrorCount = 0
 
     $validationHandler = [System.Xml.Schema.ValidationEventHandler] {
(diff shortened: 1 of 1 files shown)


logistics/scripts/modules/utility/xml-validation.psm1
```diff
@@ -77,12 +77,15 @@ function Invoke-XmlValidation {
...
 
+    $validationErrorCount = 0
+
     $validationHandler = [System.Xml.Schema.ValidationEventHandler] {
         if ($_.Severity -eq [System.Xml.Schema.XmlSeverityType]::Warning) {
             Write-Warning "$($_.Message)"
         }
         elseif ($_.Severity -eq [System.Xml.Schema.XmlSeverityType]::Error) {
             Write-Error "$($_.Message)"
+            $validationErrorCount++
```
simpat-jesus, line 88: This variable only keeps value under the validation handler scope; because of this, variable from line 80 is always 0 and the exception is never thrown, failing to stop the execution.

To get this working, set the scope to script ($script:validationErrorCount) on lines 80, 88, 122 and 123

simpat-jesus (changes requested):

logistics/scripts/modules/utility/xml-validation.psm1
mjaksn: I pushed an update with that fix

simpat-jesus (approved):
Looks good! 
image

PR #777, opened 2023-08-23
[ODS-5983] Update base image minor versions

simpat-jesus:


Docker/ods-api-web-api/alpine/mssql/Dockerfile
```diff
@@ -4,11 +4,12 @@
...
 
 # Base image with additional packages
-FROM mcr.microsoft.com/dotnet/aspnet:6.0.16-alpine3.17-amd64@sha256:dabdf423e179383d848889b760a4258a5950b224730762bf39bfd51de344cfa3 AS base
-RUN apk --no-cache add unzip=~6 dos2unix=~7 bash=~5 gettext=~0 postgresql13-client=~13 icu=~72 curl=~8
+FROM mcr.microsoft.com/dotnet/aspnet:6.0.21-alpine3.17@sha256:2ff3f278d85bd13aa630943c1e624300d9863e200f16d0cab30e09d2f3b0fc8b AS base
+RUN apk --no-cache add unzip=~6 dos2unix=~7 bash=~5 gettext=~0 icu=~72 curl=~8
 
 # Start a new layer so that the above layer can be cached
 FROM base AS build
+RUN apk --no-cache add postgresql13-client=~13
```
stephenfuqua, line 12: Why is the postgresql client needed?

simpat-jesus: Oh you are right, should be removed because this is a mssql image

simpat-jesus: Actually, we could also remove it from the pgsql images; since it is used in run.sh to wait for the pg database container to be up and running, we can use the health checks in depends_on section, and for each db container add `condition: service_healthy`

simpat-adam (approved):

PR #787, opened 2023-08-30
[ODS-5967] Fix table names and foreign key names in admin db

simpat-adam:


configuration.packages.json
```diff
@@ -1,4 +1,4 @@
-{
+﻿{
```
semalaiappan, line 1: This change not required . Please remove it .

semalaiappan (changes requested):

semalaiappan (approved):

PR #791, opened 2023-09-06
[ODS-5994] Add 7.0 Sandbox Example

simpat-jesus:


.github/workflows/Publish to Docker Hub.yml
```diff
@@ -85,7 +85,7 @@ jobs:
         shell: pwsh
 
   docker-publish:
-    if: ${{ always() && env.DOCKER_HUB_TOKEN != '' }}
```
vimayya, line 88: Don't we still need to skip publishing when DOCKER_HUB_TOKEN is not available instead of failing this auto triggered build in forks.

vimayya: DOCKER_HUB_TOKEN is being checked in FindStandardAndExtensionVersions job

semalaiappan: As Discussed , I had separate step for env.DOCKER_HUB_TOKEN  for easy understanding

semalaiappan: env.DOCKER_HUB_TOKEN != '' if condition to FindStandardAndExtensionVersions job

vimayya (approved):

PR #807, opened 2023-09-21
[ODS-6014] msql connection doesn't work with published docker images

simpat-jesus:


semalaiappan (approved):

Docker/ods-api-web-api/alpine/mssql/run.sh
```diff
@@ -11,6 +11,12 @@ if [[ "$TPDM_ENABLED" != true ]]; then
     export Plugin__Folder="./Plugin_Disabled"
 fi
 
+if [[ "$ENCRYPT_CONNECTION" == true ]]; then
```
semalaiappan, line 14: $SQLSERVER_USER" &&  "$SQLSERVER_PASSWORD  required Validation is missing here which will be handled in new ticket

PR #875, opened 2023-12-29
[ODS-6088]Cancel a running jobs on new commits

semalaiappan:


.github/workflows/Cleanup Caches by a branch.yml
```diff
@@ -9,6 +9,10 @@ on:
     branches: [main, 'ODS-*']
     types: [ closed ]
 
+concurrency:
```
simpat-jesus, line 12: This workflow executes only when a PR is closed, so there will not be an in progress workflow to cancel. This change can be reverted.

simpat-jesus (changes requested):

.github/workflows/Cleanup Caches by a branch.yml
semalaiappan: Revert it back . Thanks

simpat-jesus (approved):

vimayya (approved):

PR #878, opened 2024-01-03
[ODS-6083] Setup Ed-Fi-ODS-Implementation in Developer's Fork

semalaiappan:
github.run_number  will be different in each developer's fork Hence So package number is getting in minus number ,so added new step to reset  to 1 if gets minus for BUILD_INCREMENTER ,so that it will work in Fork also


mjaksn (approved):

mjaksn:
Could a member of @Ed-Fi-Alliance-OSS/ods-platform-automation-owners please add their approval so this can be merged.

vimayya (approved):

PR #883, opened 2024-01-09
[ODS-6116] DbDeploy update was not tracked in ODS API

JBrenesSimpat:
All packages have been updated

### DIFF
--- a/configuration.packages.json
+++ b/configuration.packages.json
@@ -2,72 +2,72 @@
   "packages": {
     "EdFiMinimalTemplate": {
       "PackageName": "EdFi.Suite3.Ods.Minimal.Template.Standard.{StandardVersion}",
-      "PackageVersion": "7.2.20",
+      "PackageVersion": "7.2.22",
       "PackageSource": "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json"
     },
     "GrandBend": {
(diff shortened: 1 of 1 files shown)


configuration.packages.json
```diff
@@ -67,7 +67,7 @@
     },
     "EdFi.Db.Deploy": {
       "PackageName": "EdFi.Suite3.Db.Deploy",
-      "PackageVersion": "4.0.31",
+      "PackageVersion": "4.0.37",
```
semalaiappan, line 70: https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-Databases/actions/runs/7463341922/job/20307788781#step:15:20 

4.0.38 is created from your branch after your change . Please update  that

configuration.packages.json
```diff
@@ -157,7 +157,7 @@
     },
     "EdFi.Ods.Common": {
       "PackageName": "EdFi.Suite3.Ods.Common",
-      "PackageVersion": "7.2.1",
+      "PackageVersion": "7.2.2",
```
semalaiappan, line 160: Why do you need to update package number here ?

configuration.packages.json
```diff
@@ -47,7 +47,7 @@
     },
     "profiles.sample": {
       "PackageName": "EdFi.Suite3.Ods.Profiles.Sample",
-      "PackageVersion": "7.2.1",
+      "PackageVersion": "7.2.11",
```
semalaiappan, line 50: Why do you need to update package number here ?

semalaiappan (changes requested):

PR #904, opened 2024-01-30
[ODS-6164] ToolsHelper require .NET 8 in PowerShell validation

JBrenesSimpat:


logistics/scripts/modules/tools/ToolsHelper.psm1
```diff
@@ -161,7 +161,7 @@ function Test-DotNetCore {
     param (
         [Parameter()]
         [int]
-        $RequiredMajor = 6,
+        $RequiredMajor = 8,
```
stephenfuqua, line 164: This function throws an exception when I run it locally, so there's something else that needs to be adjusted.

image

stephenfuqua (changes requested):

logistics/scripts/modules/tools/ToolsHelper.psm1
stephenfuqua: Sorry, I made a mistake while testing. This is working fine. 
image

stephenfuqua (approved):

PR #1088, opened 2024-07-25
[ODS-6407] Improve workflow execution

axelmarquezh:


.github/workflows/InitDev Postgres, MultiTenancy.yml
```diff
@@ -18,8 +19,6 @@ on:
         description: 'distinct ID for Trgr InitDev workflows in Implementation repo workflow'
         required: false
         default: 'distinct_id'
-  repository_dispatch:
-    types: [triggered-from-ods-repo]
```
axelmarquezh, line 22: Removed since this way of remotely triggering the workflows is no longer used.

.github/workflows/Analyze docker images.yml
```diff
@@ -8,6 +8,7 @@ name: Analyze Docker Images
 on:
   pull_request:
     branches: [main, 'b-v*-patch*','feature-*']
+    types: [opened, synchronize, reopened, unlabeled]
```
axelmarquezh, line 11: This triggers the workflow when a label gets removed from the PR.
Note that when the `type` is not specified, the default value is `opened, synchronize, reopened`.

simpat-jesus: The use of the labels is very clever to run the workflows after making a commit to an ODS PR, but I think we should only focus on the Initdev and the Dependencies Security Scan, so the default value is ok for the docker actions

simpat-jesus (changes requested):

.github/workflows/Analyze docker images.yml
axelmarquezh: I reverted the `types` definition for the Docker workflows, thanks.

simpat-jesus (approved):

semalaiappan (approved):

PR #1173, opened 2024-11-15
[ODS-6552] Change database PowerShell Installer to pack dbDeploy .nupkg instead of executable

axelmarquezh:


logistics/scripts/modules/packaging/create-package.psm1
```diff
@@ -140,6 +140,7 @@ function New-Package {
     $parameters = @()
 
     $parameters += "-p:NuspecFile=$($PackageDefinitionFile)"
+    $parameters += "-p:NoDefaultExcludes=true"
```
axelmarquezh, line 143: Without this, `dotnet pack` ignores `.nupkg` files.

PR #1177, opened 2024-11-21
[ODS-6549] Provide pre images for minor versions (7.1, 7.2)

simpat-jesus:
With this change
- Merges done to the main branch will trigger the publish of pre images. 
- Merges done to any branch different to main, will use the value from $PackageVersion to append a suffix to pre image, ex. pre-72


Docker/Invoke-Build.ps1
```diff
@@ -223,6 +223,12 @@ function Invoke-Build {
         $stdVer = ""
     }
 
+    #building the images from a branch different than main, will add the package version suffix to the pre tag
+    $preTag = "pre"
+    if ($env:BASE_BRANCH -ne 'main') {
+      $preTag = $preTag+$PackageVersion.Replace(".", "")
```
axelmarquezh, line 229: Add a `-` after the `pre` word. Right now it would generate a tag name like `pre71` which doesn't follow our existing convention.

axelmarquezh (changes requested):

axelmarquezh (approved):

semalaiappan (approved):

PR #1182, opened 2024-11-25
[ODS-6465] Create postgres role in pgsql images

semalaiappan:


axelmarquezh (changes requested):
The `bootstrap.ps1` script fails when I set `POSTGRES_USER` to a custom value in my `.env`:

image

Docker/ods-api-web-sandbox-admin/alpine/pgsql/run.sh
```diff
@@ -13,7 +13,7 @@ until PGPASSWORD=$POSTGRES_PASSWORD \
       PGHOST=$ODS_POSTGRES_HOST \
       PGPORT=$POSTGRES_PORT \
       PGUSER=$POSTGRES_USER \
-      pg_isready > /dev/null
+      pg_isready -h "$HOST" -p "$POSTGRES_PORT" -U "$POSTGRES_USER" -d "$POSTGRES_DB" > /dev/null
```
axelmarquezh, line 16: No need to specify the host, port and user since they already come from the `PGHOST`, `PGPORT`, `PGUSER` env vars respectively (a couple of lines above). So this line can be changed to:
`pg_isready -d "$POSTGRES_DB" > /dev/null`

Docker/ods-api-web-api/alpine/pgsql/run.sh
```diff
@@ -25,7 +25,7 @@ do
       PGHOST=$HOST \
       PGPORT=$POSTGRES_PORT \
       PGUSER=$POSTGRES_USER \
-      pg_isready > /dev/null
+      pg_isready -h "$HOST" -p "$POSTGRES_PORT" -U "$POSTGRES_USER" -d "$POSTGRES_DB" > /dev/null
```
axelmarquezh, line 28: No need to specify the host, port and user since they already come from the `PGHOST`, `PGPORT`, `PGUSER` env vars respectively (a couple of lines above). So this line can be changed to:
`pg_isready -d "$POSTGRES_DB" > /dev/null`

axelmarquezh (changes requested):

axelmarquezh (approved):

PR #1189, opened 2024-12-11
[ODS-6577] Verify dependent workflows are success or not- Step Added

semalaiappan:


.github/workflows/Publish to Azure Artifacts.yml
```diff
@@ -55,6 +56,49 @@ jobs:
       with:
           repository: Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation
           path: Ed-Fi-ODS-Implementation/
+    - name: Verify dependent workflows are success or not
```
axelmarquezh, line 59: This doesn't even get executed when `InitDev ... Package` fails because of the next `if` a few lines above:
```
if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch' }}
```

However,  `Docker Test` continues to get executed because this worklow technically succeeds (an skipped build is not considered a failure).

Could `ahmadnassri/action-workflow-run-wait` Github Action be useful in `Docker Test` and `Publish to Docker Hub` ?

axelmarquezh (changes requested):

.github/workflows/Publish to Azure Artifacts.yml
semalaiappan: github.event.workflow_run.conclusion == 'success'  I removed this condition . Separate step to check success for all 3 workflows before continue the next step

.github/workflows/Publish to Azure Artifacts.yml
```diff
@@ -55,6 +56,56 @@ jobs:
...
+          exit 1
+        }
+      shell: pwsh
+    # Adding a step to ensure the build job fails if it was skipped
+    - name: Fail the build job if skipped
+      if: ${{ needs.FindStandardAndExtensionVersions.result == 'skipped' }}
+      run: |
+        echo "Build job was skipped due to a failure in the previous job."
+        exit 1
+      shell: pwsh
```
axelmarquezh, line 108: These changes aren't needed since [ahmadnassri/action-workflow-run-wait](https://github.com/ahmadnassri/action-workflow-run-wait) that we use below already does these checks internally.

We just need to remove the `if` from my comment above so that the checks gets executed.

axelmarquezh (commented):
Additionally, please make the next changes to `Publish to Docker Hub` and `Docker Test` so that they don't execute 3 times every time we merge to main:
- Remove any `github.event.workflow_run.conclusion == 'success'` checks
- Add a call to `ahmadnassri/action-workflow-run-wait` in a similar way we do in this workflow

PR #1198, opened 2025-01-02
[ODS-6577] 'Publish to Docker Hub' gets executed even if 'InitDev Package' fails

axelmarquezh:


.github/workflows/Docker Test.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         with:
             timeout: 1800000
             delay: 10000
+            ignore-cancelled: true # Treat canceled as success
```
axelmarquezh, line 39: Note that the previous workflow in the chain is `Publish to Azure Artifacts` , which can be canceled during the deduplication process.
Setting `ignore-cancelled: true` makes `action-workflow-run-wait` treat cancelations as _success_, otherwise it treats them as _pending_ and times out waiting for them.

semalaiappan (approved):

PR #1207, opened 2025-01-07
[ODS-5622] Make the ODSStartup configurable 

semalaiappan:
Update EdFi.Ods.WebApi  project Appsettings as:
```json
"StartupClassName": "CustomStartup"
```

EDIT: @gmcelhanon changed the sample code to include "debug" trace statements for visibility in the debug output.

Add the `CustomStartup` class below for testing this change:

```csharp
using Autofac;
using EdFi.Ods.Api.Startup;
using EdFi.Ods.Common.Configuration;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Options;
using System.Diagnostics;

namespace EdFi.Ods.WebApi
{
    public class CustomStartup : OdsStartupBase
    {
        public CustomStartup(IWebHostEnvironment env, IConfiguration configuration)
            : base(env, configuration) { }

        public override void ConfigureServices(IServiceCollection services)
        {
            Debug.Print("CustomStartup: ConfigureServices...");
            base.ConfigureServices(services);
        }

        public override void ConfigureContainer(ContainerBuilder builder)
        {
            Debug.Print("CustomStartup: ConfigureContainer...");
            base.ConfigureContainer(builder);
        }
        
        public override void Configure(
            IApplicationBuilder app,
            IWebHostEnvironment env,
            IOptions<ApiSettings> apiSettingsOptions,
            IApplicationConfigurationActivity[] configurationActivities)
        {
            Debug.Print("CustomStartup: Configure...");
            base.Configure(app, env, apiSettingsOptions, configurationActivities);
        }
    }
}
```


Application/EdFi.Ods.WebApi/Program.cs
```diff
@@ -37,7 +38,7 @@ public static async Task Main(string[] args)
                         webBuilder.ConfigureKestrel(
                             serverOptions => { serverOptions.AddServerHeader = false; });
 
-                        webBuilder.UseStartup<Startup>();
+                        webBuilder.UseStartup<OdsStartupBase>(context => StartupFactory.GetStartup(context));
```
gmcelhanon, line 41: Look at Line 26 above. You could add a line after that like the one below to read the StartupClassName setting into a variable which is then used to drive the logic in this section:
```csharp
string startupClassName = config.GetValue<string>("StartupClassName");
```
With a value of "Startup" or a call to `string.IsNullOrEmpty` returning `true`, you would then just use the original code that was here:
```
webBuilder.UseStartup<Startup>();
```
With any other value, you'd then proceed down the path of logic you have, but I'd suggest skipping the rebuild of the configuration there (currently at line 24 of _StartupFactory.cs_), preferring instead an approach making the `StartupFactory` class non-static, and passing the startup class name value to its constructor.

Then, the `webBuilder.UseStartup` method overload that accepts a factory method is passed the public instance method rather than a static method, and all the logic related to building the configuration and extracting the "StartupClassName" setting is removed from the factory.

Here's my suggestion for the first part of the `Main` method:
```csharp
string startupClassName;

using (var hostBuilderInitial = Host.CreateDefaultBuilder(args).Build())
{
    var config = hostBuilderInitial.Services.GetService<IConfiguration>();
    startupClassName = config.GetValue<string>("StartupClassName");

    AssemblyLoaderHelper.LoadAssembliesFromFolder(
        AssemblyLoaderHelper.GetPluginFolder(
            config?.GetValue<string>("Plugin:Folder") ?? string.Empty));
}

var hostBuilder = Host.CreateDefaultBuilder(args)
    .ConfigureLogging(ConfigureLogging)
    .UseServiceProviderFactory(new AutofacServiceProviderFactory())
    .ConfigureWebHostDefaults(
        webBuilder =>
        {
            webBuilder.ConfigureKestrel(
                serverOptions => { serverOptions.AddServerHeader = false; });

            if (string.IsNullOrEmpty(startupClassName) || startupClassName.EqualsIgnoreCase("Startup"))
            {
                webBuilder.UseStartup<Startup>();
            }
            else
            {
                var startupFactory = new StartupFactory(startupClassName);
                webBuilder.UseStartup(context => startupFactory.Create(context));
            }
        });
```
And the instance-based startup factory:
```csharp
public class StartupFactory
{
    private readonly string _startupClassName;

    public StartupFactory(string startupClassName)
    {
        _startupClassName = startupClassName;
    }

    public OdsStartupBase Create(WebHostBuilderContext arg)
    {
        // Scanning logic here based on _startupClassName
    }
}
```

Application/EdFi.Ods.WebApi/StartupFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+            OdsStartupBase startup = (OdsStartupBase)Activator.CreateInstance(type, webHostBuilderContext.HostingEnvironment, webHostBuilderContext.Configuration);
+            return startup;
+        }
+
+        private static Type FindStartupClassType(string startupClassName)
+        {
+            var startupType = AppDomain.CurrentDomain.GetAssemblies()
+                .SelectMany(assembly => assembly.GetTypes())
+                .FirstOrDefault(t => t.Name == startupClassName);
+
```
gmcelhanon, line 42: Because the return type is expected to inherit from `OdsStartupBase`, you should also filter the types for whether they are assignable to `OdsStartupBase`.

Application/EdFi.Ods.WebApi/StartupFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.IO;
+
+namespace EdFi.Ods.WebApi
+{
+    public static class StartupFactory
+    {
+       
+        private const string MissingStartupClassMessage =
+            "Could not find the startup class. No loaded class member could be found matching the configured startup class name '{0}'.";
```
gmcelhanon, line 19: Change this message to:
`Could not find a startup class named '{0}' that derives from OdsStartupBase.`

Application/EdFi.Ods.WebApi/appsettings.json
```diff
@@ -12,6 +12,7 @@
     },
     "Tenants": {
     },
+    "StartUpClassName": "Startup",
```
gmcelhanon, line 15: Rename this setting to "StartupClassName", and default it to an empty string (which should cause the existing `Startup` class to be used by default, as described above). This also reduces the chance of this change breaking someone who inadvertently misses the setting in their configuration.

gmcelhanon (changes requested):

Application/EdFi.Ods.WebApi/StartupFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+            var startup = (OdsStartupBase)Activator.CreateInstance(startupType, webHostBuilderContext.Configuration);
+            return startup;
+        }
+
+        private Type FindStartupClassType(string startupClassName)
+        {
+            var startupType = AppDomain.CurrentDomain.GetAssemblies()
+                .Where(assembly => assembly.GetName().Name?.StartsWith("EdFi.") ?? false)
+                .SelectMany(assembly => assembly.GetTypes())
+                .Where(t => typeof(OdsStartupBase).IsAssignableFrom(t) && t.Name == startupClassName)
```
gmcelhanon, line 31: Without digging, I'm not sure how expensive the `IsAssignableFrom` method is, but I know that string comparison should be very fast. Because of C#'s behavior in short circuiting conditional logic once part of the criteria is false, perform the string check first, and change it to use the `EqualsIgnoreCase` method while you're at it. So this line becomes:
```
.Where(t => t.Name.EqualsIgnoreCase(startupClassName) && typeof(OdsStartupBase).IsAssignableFrom(t))
```

gmcelhanon (changes requested):
Looks pretty good now ... just one more tweak.

Application/EdFi.Ods.WebApi/StartupFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+    public class StartupFactory(string startupClassName)
+    {
+        private readonly string _startupClassName = startupClassName;
+        private const string MissingStartupClassMessage =
+            "Could not find a startup class named '{0}' that derives from OdsStartupBase.";
+
+        public OdsStartupBase Create(WebHostBuilderContext webHostBuilderContext)
+        {
+            var startupType = FindStartupClassType(_startupClassName);
+            var startup = (OdsStartupBase)Activator.CreateInstance(startupType, webHostBuilderContext.Configuration);
```
gmcelhanon, line 23: In testing, this is missing an argument from the custom startup class' constructor, which we can easily pass along.

Updated code should appear as follows:
```csharp
            var startup = (OdsStartupBase)Activator.CreateInstance(
                startupType,
                webHostBuilderContext.HostingEnvironment,
                webHostBuilderContext.Configuration);
```

Application/EdFi.Ods.WebApi/StartupFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+            return startup;
+        }
+
+        private Type FindStartupClassType(string startupClassName)
+        {
+            var startupType = AppDomain.CurrentDomain.GetAssemblies()
+                .Where(assembly => assembly.GetName().Name?.StartsWith("EdFi.") ?? false)
+                .SelectMany(assembly => assembly.GetTypes())
+                .Where(t => t.Name.EqualsIgnoreCase(startupClassName) && typeof(OdsStartupBase).IsAssignableFrom(t))
+                .FirstOrDefault();
```
gmcelhanon, line 33: While we're in this file, simplify the statement by replacing the use of `Where` with `FirstOrDefault` (rather than repeating it at the end) since `FirstOrDefault` also accepts a predicate (so the `Where` can be eliminated).

Application/EdFi.Ods.WebApi/StartupFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            if (startupType == null)
+            {
+                throw new TypeLoadException(string.Format(MissingStartupClassMessage, startupClassName));
+            }
+            return startupType;
+        }
+
```
gmcelhanon, line 41: Remove this blank line.

Application/EdFi.Ods.WebApi/appsettings.json
```diff
@@ -12,6 +12,7 @@
     },
     "Tenants": {
     },
+    "StartUpClassName": "",
```
gmcelhanon, line 15: Change the capitalization of the name in this setting to `StartupClassName` (Startup is treated as one word, not two).

gmcelhanon (changes requested):

gmcelhanon (approved):

PR #1215, opened 2025-01-16
[ODS-6613] Deserialization fails when a model extension introduces an entity extension after the resource has been created

gmcelhanon:


gmcelhanon:
CLOSE - DO NOT MERGE

PR #1235, opened 2025-04-01
[ODS-6639] Sandbox Admin for Certification Testing

mjaksn:
…ma seperation

### DIFF
--- a/Application/EdFi.Ods.SandboxAdmin/wwwroot/js/application.js
+++ b/Application/EdFi.Ods.SandboxAdmin/wwwroot/js/application.js
@@ -109,7 +109,7 @@ var addApplicationDialog = function () {
         },
         pattern: {
             message: 'Please enter a valid number with at least six digits of educationOrganizationId',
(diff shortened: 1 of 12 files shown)


Application/EdFi.Ods.SandboxAdmin/wwwroot/js/application.js
```diff
@@ -109,7 +109,7 @@ var addApplicationDialog = function () {
         },
         pattern: {
             message: 'Please enter a valid number with at least six digits of educationOrganizationId',
-            params: /^\d{6,}$/
+            params: /^(?:\s*\d{6,18},?)+$/
```
mjaksn, line 112: This regex should validate the user input with one or more 6-18 digit EdOrgIds separated with commas, and allow whitespace in front of eachEdOrgId (this is so that spaces after commas, which will be trimmed away during parsing, are allowed in the input - Example: `123456, 555555, 666666` is valid).

simpat-jesus (approved):

PR #1271, opened 2025-09-09
[ODS-6673] Pin `Docker Test` and `Publish to Docker` to the `b-v7.3-patch1` version

axelmarquezh:


.github/workflows/Publish to Azure Artifacts.yml
```diff
@@ -120,14 +120,14 @@ jobs:
...
-    uses: Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/.github/workflows/Docker Test.yml@main
+    uses: Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/.github/workflows/Docker Test.yml@2e0518d04301f599c29ca53f7a79436e28e5a1b2
     with:
       initdev_commit: ${{ github.event.workflow_run.head_sha || github.sha }}
       initdev_head_branch: ${{ github.event.workflow_run.head_branch || github.head_ref || github.ref_name }}
       
   PublishToDockerHub:
     needs: DockerTest
-    uses: Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/.github/workflows/Publish to Docker Hub.yml@main
+    uses: Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/.github/workflows/Publish to Docker Hub.yml@2e0518d04301f599c29ca53f7a79436e28e5a1b2
```
axelmarquezh, line 130: This commit hash points to the `b-v7.3-patch1` branch.
The change is needed because, in main, these workflows set `CURRENT_STANDARD_VERSION: "6.0.0"` ([here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/81142d5580431366f3a6937dc7608f926b453cd8/.github/workflows/Publish%20to%20Docker%20Hub.yml#L37)); the variable is used to know when to add the DS version suffix to the Docker tags, for example, `pre-73` vs `pre-73-5.2.0`. If the current DS version is `5.2.0`, then we should not add the suffix (per our convention).

PR #1278, opened 2025-09-12
[ODS-6673] Improve Docker workflows trigger logic (#1272)

semalaiappan:


Docker/get-versions.ps1
```diff
@@ -129,7 +129,12 @@ function Get-NugetPackageVersion {
 
         # Enable usage of prereleases
         [Switch]
-        $PreRelease
+        $PreRelease,
+
+        # Search for a package whose patch version is lower than or equal to the given value.
+        # Useful, for example, to not select a v7.3.1 package from a v7.3 context
+        [int]
+        $MaxPatchVersion = 9999
```
axelmarquezh, line 137: Replace this file's contents with the `main` version, but keep `$PackageVersion = "7.1"` and ` [ValidateSet('4.0.0', '5.0.0')]`.

Docker/Invoke-Build.ps1
```diff
@@ -131,7 +131,7 @@ param (
     # Major/minor version number
     [Parameter()]
     [string]
-    $PackageVersion = "7.3",
+    $PackageVersion = "7.1",
```
axelmarquezh, line 134: Replace this file's contents with the `main` version, but keep `$PackageVersion = "7.1"`.

axelmarquezh (changes requested):
The reasoning behind these comments is that we used to execute the main definition of these workflows (when triggered by `workflow_run`), but we recently changed the philosophy to execute the patch's definition instead, meaning that we need to bring the `main` definition of the workflows with the appropriate updates to handle the correct DS version and ODS version.

PR #1282, opened 2025-09-19
[ODS-4330] EdFi.Suite3.Installer.WebApi EdFi.Suite3.Installer.SwaggerUI EdFi.Suite3.Installer.SandboxAdmin Issue CodeFix

semalaiappan:


logistics/scripts/modules/packaging/create-package.psm1
```diff
@@ -157,16 +156,32 @@ function New-Package {
     $parameters += "--configuration"
     $parameters += $BuildConfiguration
 
-    $nuspecProperties = "-p:NuspecProperties=""version=$($Version)"
 
 
-    foreach ($prop in $Properties) {
-        $nuspecProperties += ";$prop"
+    [xml]$xml = Get-Content -Path $PackageDefinitionFile
```
semalaiappan, line 161: Looping each key/value and updating it in nuget spec file with value here .  Currently dotnet pack is taking values

logistics/scripts/modules/packaging/create-package.psm1
```diff
@@ -214,10 +229,6 @@ function Publish-PrereleasePackage {
         "--api-key", $ApiKey
     )
 
-    if ($Verbose) {
```
semalaiappan, line 217: This debugging is not required here ,so removed

logistics/scripts/modules/packaging/create-package.psm1
vimayya: Should we reuse New-Nuspec https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/b-v7.1-patch5/logistics/scripts/modules/packaging/packaging.psm1#L8 here?

axelmarquezh (approved):

logistics/scripts/modules/packaging/create-package.psm1
axelmarquezh: Rather than creating a new .nuspec file, he's overwriting some values from the existing .nuspec.

He's now setting the `copyright` and `version` fields (that come from the `$Properties` param) in the .nuspec file because, when specified in pack's command line argument, the comma in the license was causing issues ([here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/actions/runs/17805969224/job/50617695373#step:4:26), for example).

PR #1392, opened 2026-06-23
[ODS-6847] Redis appsettings: adding CachingMode for hybrid mode

adiaz-msdf:
### Summary

- Added the CachingMode for the appSettings to enable the "Hybrid" | "External" cache mode.
- Added property{CorrelationId} for the logs displays the key id when fails

### DIFF
--- a/Application/EdFi.Ods.WebApi/log4net.config
+++ b/Application/EdFi.Ods.WebApi/log4net.config
@@ -3,7 +3,7 @@
   <appender name="TraceAppender" type="log4net.Appender.TraceAppender">
     <threshold value="INFO" />
     <layout type="log4net.Layout.PatternLayout">
-      <conversionPattern value="%date [%thread,%property{ApiClientId}] %-5level %logger{1}.%method [%line] - MESSAGE: %message%newline %exception" />
+      <conversionPattern value="%date [%thread,%property{ApiClientId},%property{CorrelationId}] %-5level %logger{1}.%method [%line] - MESSAGE: %message%newline %exception" />
     </layout>
   </appender>
   <appender name="RollingFile" type="log4net.Appender.RollingFileAppender">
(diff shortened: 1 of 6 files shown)


Application/EdFi.Ods.WebApi/log4net.config
```diff
@@ -3,7 +3,7 @@
   <appender name="TraceAppender" type="log4net.Appender.TraceAppender">
     <threshold value="INFO" />
     <layout type="log4net.Layout.PatternLayout">
-      <conversionPattern value="%date [%thread,%property{ApiClientId}] %-5level %logger{1}.%method [%line] - MESSAGE: %message%newline %exception" />
+      <conversionPattern value="%date [%thread,%property{ApiClientId},%property{CorrelationId}] %-5level %logger{1}.%method [%line] - MESSAGE: %message%newline %exception" />
```
stephenfuqua, line 13: Many of our customers already have patterns in place for parsing log entries, and they do not like surprises when something changes.

It will be best if we provide this as an example and document it, but do not change the default log4net configuration.

For example, we might provide both `ConsoleAppender` and `ConsoleAppenderWithCorelationId`, then modify the `root` element below to include something like this:

```
<!-- To include Correlation ID in the output log, uncomment the follow line, and comment out the ConsoleAppender line above -->
<!-- appender-ref ref="ConsoleAppenderWithCorelationId" / -->
```

In this case, we would also simply revert the changes to `TraceAppender` and `ConsoleAppender`.

Application/EdFi.Ods.WebApi/log4net.development.config
```diff
@@ -21,7 +21,7 @@
   <appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender">
     <threshold value="DEBUG" />
     <layout type="log4net.Layout.PatternLayout">
-      <conversionPattern value="%date [%thread,%property{ApiClientId}] %-5level %logger{1} - %message%newline" />
+      <conversionPattern value="%date [%thread,%property{ApiClientId},%property{CorrelationId}] %-5level %logger{1} - %message%newline" />
```
stephenfuqua, line 24: Because this is a `development` config, leaving the Correlation Id here is probably acceptable.

Application/EdFi.Ods.WebApi/log4net.config
adiaz-msdf: Fixed

stephenfuqua (approved):

